PR #9, opened 2024-04-16
feat: expose decimal and ipaddr constructors

### LINKED ISSUE #8: Expose `newDecimalValue` Function for Simplified Decimal Conversion
**Description:**

The `cedar-go` library uses an `int64` data type to represent decimal values, necessitating specific conversion logic to translate string representations of decimals into their numeric counterparts. Currently, this conversion is only accessible through the `Decimal.UnmarshalJSON` method.

In practice, this implementation poses limitations, especially when integrating with systems that do not use JSON for serialization, such as gRPC. For instance, when receiving data in formats other than the expected JSON structure (`{"__extn": {"fn": "decimal", "arg": "12.5"}}`), there is no straightforward way to utilize `cedar-go`'s conversion logic directly.

To enhance flexibility and usability, I propose making the `newDecimalValue` function publicly accessible. This change would allow developers to directly convert a decimal string like `"12.5"` into the `cedar.Decimal` format (e.g., `cedar.Decimal(125000)`) without dealing with JSON-specific constraints.

Such an enhancement would greatly benefit those of us integrating `cedar-go` with various RPC frameworks and would streamline the handling of decimal data across different platforms.

**Impact:**

- **Broader Usability:** Makes `cedar-go` more versatile and easier to integrate with non-JSON based systems.
- **Simplified Development:** Reduces the need for custom conversion functions or workarounds, leading to cleaner, more maintainable code.

**Proposed Solution:**

Expose the `newDecimalValue` function as a public method, thereby extending the utility of the `cedar-go` library to handle decimal conversions more flexibly.

> If you see fit, I can send a PR making this change.

caiorcferreira:
Closes #8 

*Description of changes:*

Expose `newDecimalValue` and `newIPAddr` functions as public methods.


philhassey:
@caiorcferreira Thanks for the PR!  I see in the PR that this re-did all the indenting of a bunch of files.  Can you set your indent settings to the same as this repo and update the PR?  Thanks!

caiorcferreira:
@philhassey I think now it has the original indenting.

philhassey:
@caiorcferreira Thanks!  Can you add a doc string to the two new functions?

caiorcferreira:
@philhassey just added!

For curiosity, why represent decimals using `int64`? Cedar Rust also does it?

value.go
```diff
@@ -405,7 +405,11 @@ type Decimal int64
 // DecimalPrecision is the precision of a Decimal.
 const DecimalPrecision = 10000
 
-func newDecimalValue(s string) (Decimal, error) {
+// ParseDecimal takes a string representation of a decimal number and converts it into a Decimal type.
```
philhassey, line 408: I actually think just the first line is the only one that is necessary.  (Thanks for all the detail though, but it seems a bit much to me!)

philhassey:
@caiorcferreira Thanks for the docs, I suggested shortening them, I think that'll be easier to read.

I don't recall if int64 is the underlying type in Rust, but if you look at the docs:
https://docs.cedarpolicy.com/policies/syntax-datatypes.html#datatype-decimal

It seems like that's the range that this was made for.

kjamieson-sdm (commented):
PR looks good to me as well, although I agree with @philhassey that the function description comments are probably a little unnecessarily verbose, given we're just following the spec there.

philhassey (approved):
Looks good!

kjamieson-sdm (approved):

philhassey:
Thanks @caiorcferreira !

Issue #17, opened 2024-06-12
Change Go mod spec to 1.22 from 1.22.0


grsubramanian:
We are in the process of incorporating cedar-go into our product.

And we are hitting a snag. Currently, cedar-go specifies a minimum Go version of 1.22.0 in its go.mod file.

Go 1.22.0 seems to have several security vulnerabilities that have since been fixed in higher Go 1.22 versions.

Is it possible to change the Go version for cedar-go in go.mod to 1.22, instead of 1.22.0?

200sc:
See #18 for discussion.

PR #18, opened 2024-06-12
Set Go 1.22 as min Go version

grsubramanian:
*Issue #, if available:* https://github.com/cedar-policy/cedar-go/issues/17

*Description of changes:* Sets Go 1.22 as the minimum Go version.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -1,5 +1,5 @@
 module github.com/cedar-policy/cedar-go
 
-go 1.22.0
+go 1.22
 
 require golang.org/x/exp v0.0.0-20240222234643-814bf88cf225


philhassey (commented):
Looks good to me.

200sc:
@grsubramanian Specifying a minimum of 1.22 would still allow users to use 1.22.0; and specifying 1.22.0 does not prevent users from using e.g. 1.22.3. 

Do you have an example of a workflow this enables that doesn't work for you today?

grsubramanian:
> Specifying a minimum of 1.22 would still allow users to use 1.22.0

Absolutely. Am not trying to prevent anyone from using 1.22.0.

> and specifying 1.22.0 does not prevent users from using e.g. 1.22.3.

Hmm. When I add a dependency on this lib via `go get github.com/cedar-policy/cedar-go`, it changes our go directive from 1.22 to 1.22.0, and adds a toolchain directive specifying 1.22.3. As you say, one could very well use the toolchain directive to pick a different Go version.

That said, our vulnerabilities scanner (Snyk specifically) flags issues with the use of 1.22.0 in the Go directive. There are quite a few vulnerabilities, and several quite severe ones. This makes it hard for us to ignore / suppress the flagged issues.

> Do you have an example of a workflow this enables that doesn't work for you today?

With the change in this PR, we are able to run `go get github.com/cedar-policy/cedar-go` with no changes to our go.mod except the obvious added require directive. And our vulnerability scanner finds this likeable and reports no issues.

200sc:
@grsubramanian If you specify `go 1.22.3` in your go.mod (not `toolchain 1.22.3`), `go mod tidy` should (per my testing) combine `go 1.22.3` with any same versioned toolchain line, and would in theory remove those warnings. And running `go get github.com/cedar-policy/cedar-go` after should not change that 1.22.3 pin.

grsubramanian:
> If you specify go 1.22.3 in your go.mod (not toolchain 1.22.3), go mod tidy should (per my testing) combine go 1.22.3 with any same versioned toolchain line, and would in theory remove those warnings. And running go get github.com/cedar-policy/cedar-go after should not change that 1.22.3 pin.

That was my observation as well. But our team tends to prefer to not pin to a specific patch version (e.g. 1.22.3) but rather pin to a specific minor version (e.g. 1.22). Gives us more flexibility without expressing the intent that the software is only meant to work for a particular patch version.

For the health of the Cedar go library too, it would be nice it if didn't pin to a specific patch version (especially 1.22.0, which has security issues), but pinned to just a minor version.

200sc:
I think there's a misunderstanding (perhaps on snyk's part); the use of 1.22.0 doesn't suggest you must use that version, it defines a minimum. Declaring 1.22 is equivalent to declaring 1.22.0. 

If the goal is to avoid security problems in a package consuming this library, that package should specify 1.22.3 as it's minimum version. 1.22 and 1.22.0 are exactly equivalent from this library's perspective on security. If we want to improve the security of this package, we must specify 1.22.3 instead. I would approve a PR that did that, although I do not believe it is necessary as client applications can do this themselves.

200sc:
So the only tangible benefit to using 1.22 over 1.22.0, would be to appease security tooling which is misunderstanding what that directive means; if snyk was a perfect actor it would equally reject 1.22 and 1.22.0. However, if enough users are using such tooling and find manually adjusting their go.mod files annoying, we could also accept this PR to improve the usability (not the security) of the library.

200sc:
Also interesting, if we're thinking of security: specifying `go 1.22` would enable the use of `go 1.22rc1` where it is currently forbidden, so, if the security of consuming applications was this library's concern, it  would in theory be actually less secure to specify `go 1.22`.

200sc:
So all of that said; if we agree this is for convenience and is less secure than the status quo, I would approve merging this PR. (or, I'm happy to be proven wrong on anything I said).

grsubramanian:
All of what you've said makes sense Patrick (200sc). And I did have some misunderstanding about Go versioning, which you've kindly cleared up.

This is really more about convenience. In our go.mod file, we set Go version to 1.22, and at runtime, pick the latest patch version available. This setup is convenient because we don't have to keep adjusting the patch version in our Go mod file each time a new vulnerability has been discovered with a Go patch version.

Setting 1.22.0 in cedar-go's go.mod file forces us to set a patch version (1.22.0 or higher) in our go.mod file, whereas setting 1.22 in cedar-go's go.mod file allows to do all that, but also allows us to set 1.22 in our go.mod file, which as explained above is more convenient for us.

It turns out that setting Go to 1.22.3 also causes Snyk to complain. We are investigating this.

If it is all the same to you, I'd appreciate if we can move forward with this PR, as it helps unblock our use of Cedar.

PR #16, opened 2024-06-04
Fix corpus tests

khieta:
*Issue #, if available:*

*Description of changes:*

Hi 👋 I've been working on standardizing Cedar's JSON interfaces (including the integration tests), and made a couple edits recently that will affect `cedar-go`. I believe the changes I've made here will fix any breaks.

Tested with: `go test -count=1 -v -tags corpus ./integration_tests/...`

Related PRs:
* Update to `cedar`: https://github.com/cedar-policy/cedar/pull/837
* Update to `cedar-integration-tests`: https://github.com/cedar-policy/cedar-integration-tests/pull/4

PR #22, opened 2024-07-03
Fix up nightly corpus tests

patjakdev:
Issue #, if available: None

Description of changes:

This PR causes the Github action to fail correctly when the corpus tests have failed such that our alerting will actually be invoked.

I've also pulled in the fixes from https://github.com/cedar-policy/cedar-go/pull/16 so that the tests now pass.

I'd like to make additional changes to get these to run on every PR and to check in a copy of the tests to this repo and change the alerting to let us know when the tests are out of sync with the upstream version, but I'll save those for later. This at least gets us back to status quo.

### DIFF
--- a/.github/workflows/corpus.yml
+++ b/.github/workflows/corpus.yml
@@ -7,7 +7,11 @@ on:
 
 jobs:
   build:
+    defaults:
+      run:
+        shell: bash
     runs-on: ubuntu-latest
+
     steps:
       - uses: actions/checkout@v4
 
--- a/.github/workflows/corpus.yml
+++ b/.github/workflows/corpus.yml
@@ -30,7 +30,7 @@ jobs:
         run: go test -count=1 -v -tags corpus ./integration_tests/... | tail
 
       - name: Notify on Failure
-        if: failure()
+        if: failure() && github.event_name == 'schedule'
         uses: actions/github-script@v6
         with:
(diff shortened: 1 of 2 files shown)


.github/workflows/corpus.yml
```diff
@@ -7,7 +7,11 @@ on:
 
 jobs:
   build:
+    defaults:
+      run:
+        shell: bash
```
patjakdev, line 12: Setting `bash` here invokes the default `-o pipefail` behavior that correctly makes it so that the `go test -count=1 -v -tags corpus ./integration_tests/... | tail` fails when the tests fail.

.github/workflows/corpus.yml
```diff
@@ -26,7 +30,7 @@ jobs:
         run: go test -count=1 -v -tags corpus ./integration_tests/... | tail
 
       - name: Notify on Failure
-        if: failure()
+        if: failure() && github.event_name == 'schedule'
```
patjakdev, line 33: We only want to be notified when this fails in the nightly run. Manual invocations shouldn't file an issue.

philhassey (approved):
Short of a manual test of this here, it looks good.

patjakdev:
> Short of a manual test of this here, it looks good.

You can see a successful run of the tests over in the [StrongDM fork](https://github.com/strongdm/cedar-go/actions/runs/9785686649).

PR #28, opened 2024-09-05
cedar: change JSON marshaling of the Position struct to use conventional lower case keys

patjakdev:
*Issue #, if available:* None

*Description of changes:*

Technically, this is a breaking change and will require us to go to 0.3.0.

### DIFF
--- a/policy.go
+++ b/policy.go
@@ -97,10 +97,17 @@ func (p *Policy) Effect() Effect {
 
 // A Position describes an arbitrary source position including the file, line, and column location.
 type Position struct {
-	Filename string // optional name of the source file for the enclosing policy, "" if the source is unknown or not a named file
-	Offset   int    // byte offset, starting at 0
-	Line     int    // line number, starting at 1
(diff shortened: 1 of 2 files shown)


philhassey:
Can you adjust the doc comments as per the Go recommendations?

https://tip.golang.org/doc/comment#:~:text=type%20Buffer%20struct%20%7B%0A%20%20%20%20...%0A%7D-,For,-a%20struct%20with

(the writing style you have would be post-attribute.  For pre-attribute it should include the attribute name as the first word.)

philhassey (approved):
Approved, though the doc fix would be nice.

patjakdev:
> Can you adjust the doc comments as per the Go recommendations?
> 
> https://tip.golang.org/doc/comment#:~:text=type%20Buffer%20struct%20%7B%0A%20%20%20%20...%0A%7D-,For,-a%20struct%20with
> 
> (the writing style you have would be post-attribute. For pre-attribute it should include the attribute name as the first word.)

@philhassey Done! Thanks for catching that.

PR #29, opened 2024-09-06
This commit adds an extension type, datetime, to Cedar Go.

apg:
See _pending_ [RFC 80](https://github.com/strongdm/cedar-rfcs/blob/datetime-rfc/text/0080-datetime-extension.md) for details.

This PR does not add the duration type, nor the methods `toTime()`, `offset()`, or `durationSince()` that would require it.

This PR adds operator overloading, by introducing a set of "Virtual" comparison Evaler. Types that can overload operators must implement the `Lesser` interface. Long and Datetime are implemented in this PR. (Note: It is expected that there will be a Decimal operator overload RFC at some later date, but that is not hooked into this)

The now obsolete Evalers for Long comparisons have not been removed. A possible improvement to this PR would be to utilize them when it is known that Longs are on the right and left side.

The datetime itself is backed by an int64, and can be lifted from a Go `time.Time`, integer or float "unsafely."

### DIFF
--- a/types/datetime.go
+++ b/types/datetime.go
@@ -171,7 +171,6 @@ func ParseDatetime(s string) (Datetime, error) {
 		}
 	}
 
-	fmt.Printf("WE'RE HERE: %s\n", s[i:length])
 
 	if i >= length {
 		return Datetime{}, fmt.Errorf("%w: expected timezone indicator", ErrDatetime)
--- a/types/datetime.go
+++ b/types/datetime.go
@@ -2,6 +2,7 @@ package types
 
 import (
 	"encoding/json"
+	"errors"
 	"fmt"
 	"strconv"
 	"time"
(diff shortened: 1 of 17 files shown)


types/datetime.go
```diff
@@ -0,0 +1,267 @@
...
+			}
+
+			milli = 100*int(rune(s[i])-'0') + 10*int(rune(s[i+1])-'0') + int(rune(s[i+2])-'0')
+			i = i + 3
+		} else {
+			return Datetime{}, fmt.Errorf("%w: invalid millisecond", ErrDatetime)
+		}
+	}
+
+	fmt.Printf("WE'RE HERE: %s\n", s[i:length])
```
apg, line 174:
```suggestion
```

types/datetime.go
```diff
@@ -0,0 +1,266 @@
...
+	return []byte(`datetime("` + a.String() + `")`)
+}
+
+func (a Datetime) String() string {
+	return time.UnixMilli(a.Value).UTC().Format("2006-01-02T15:04:05.000Z")
+}
+
+func (a Datetime) UnmarshalJSON(b []byte) error {
+	// TODO
+	return nil
```
apg, line 250: Apparently got a bit ahead of myself. Implementing now.

apg:
Whoops. I meant to open this on the strongdm/cedar-go branch. Closing.

PR #32, opened 2024-09-11
Several improvements to the upstream corpus test workflow

patjakdev:
*Issue #, if available:* None

*Description of changes:*

* Only run this workflow on the official cedar-go repo
* Update `actions/github-script` to @v7 to avoid warnings about using an old NodeJS version
* Only open an issue if the compare step failed. We don't necessarily want to open an issue if the `curl` fails for some reason.
* Give the job permission to create issues
* Only open an issue if an open one does not already exist


philhassey (approved):
LGTM

PR #33, opened 2024-09-13
Fix parsing of reserved keywords

patjakdev:
*Issue #, if available:* None

*Description of changes:*

This PR modifies the Cedar tokenizer to properly reserve the set of [reserved keywords](https://arc.net/l/quote/yvyhkfvf) from the Cedar spec. In addition, it resolves a parsing ambiguity for the `Primary` production, which requires a lookahead to determine whether it will be a `VAR`, `Entity`, or `ExtFun`.

### DIFF
--- a/internal/parser/cedar_unmarshal_test.go
+++ b/internal/parser/cedar_unmarshal_test.go
@@ -553,6 +553,11 @@ func TestParseApproximateErrors(t *testing.T) {
 		{"func", `permit (principal, action, resource) when { ip(}`, "invalid primary"},
 		{"args", `permit (principal, action, resource) when { ip(42 42)`, "got 42 want ,"},
 		{"dupeKey", `permit (principal, action, resource) when { {k:42,k:43}`, "duplicate key"},
+		{"reservedKeywordAsRecordKey", `permit (principal, action, resource) when { {false:43} }`, "expected ident or string"},
+		{"reservedKeywordAsHas", `permit (principal, action, resource) when { {} has false }`, "expected ident or string"},
+		{"reservedKeywordAsEntityType", `permit (principal == false::"42", action, resource)`, "expected ident"},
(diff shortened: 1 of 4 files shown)


patjakdev:
See further discussion of this PR on the [StrongDM fork](https://github.com/strongdm/cedar-go/pull/7)

kjamieson-sdm:
It looks like this forbids the use of reserved keywords in annotation keys, which the Rust implementation specifically allows (see https://github.com/cedar-policy/cedar/pull/634). Are we concerned about that difference in behaviour?

internal/parser/cedar_unmarshal_test.go
```diff
@@ -584,3 +590,96 @@ func TestPolicySliceErrors(t *testing.T) {
...
+			nil, testutil.Error}, // Rust: unexpected token
+		{"scope/unless", `permit (principal == unless::"test", action, resource);`,
+			ast.Permit().PrincipalEq(types.NewEntityUID("unless", "test")), testutil.OK}, // Rust: OK
+		{"scope/when", `permit (principal == when::"test", action, resource);`,
+			ast.Permit().PrincipalEq(types.NewEntityUID("when", "test")), testutil.OK}, // Rust: OK
+	}
+	for _, tt := range tests {
+		tt := tt
+		t.Run(tt.name, func(t *testing.T) {
+
```
kjamieson-sdm, line 675: Is `t.Parallel()` omitted intentionally?

patjakdev:
> It looks like this forbids the use of reserved keywords in annotation keys, which the Rust implementation specifically allows (see [cedar-policy/cedar#634](https://github.com/cedar-policy/cedar/pull/634)). Are we concerned about that difference in behaviour?

Good catch @kjamieson-sdm. I was basing this on the [documented grammar](https://arc.net/l/quote/hzojfyne).

I'll add a test specifically for this case.

kjamieson-sdm (approved):

PR #34, opened 2024-09-13
General performance improvements and experimental batch mode

patjakdev:
*Issue #, if available:* None

*Description of changes:*

This PR addresses two issues:
1. The bulk of the changes are an experimental batched mode for making authorization requests against a policy set. This is similar in spirit to partial evaluation, but actually calculates the permutations of the various unknowns given a set of possible values for each unknown and returns an "allow" or "deny" decision for each permutation. It also allows for certain portions of the request to be ignored when evaluating policy, which allows for answering questions like "What resources might Bob be able to access given any possible context?", where the set of values cannot be known ahead of time.
2. A performance problem when traversing a large, flat entity graph. This is addressed by lazy allocating the entity the parent cache only when an entity has parents.


patjakdev:
See more discussion about this PR on the [StrongDM fork](https://github.com/strongdm/cedar-go/pull/6)

types/authorize.go
```diff
@@ -0,0 +1,92 @@
...
+	return nil
+}
+
+// A Diagnostic details the errors and reasons for an authorization decision.
+type Diagnostic struct {
+	Reasons []DiagnosticReason `json:"reasons,omitempty"`
+	Errors  []DiagnosticError  `json:"errors,omitempty"`
+}
+
+// A DiagnosticReason details the Policy index within a PolicySet, and the Position within
```
kjamieson-sdm, line 43: I realize this code was just moved from elsewhere, but would it more correct to change "Policy index" to "Policy identifier" or "PolicyID" since I think it is no longer an index?

x/exp/batch/batch.go
```diff
@@ -0,0 +1,373 @@
...
+
+var errUnboundVariable = fmt.Errorf("unbound variable")
+var errUnusedVariable = fmt.Errorf("unused variable")
+var errMissingPart = fmt.Errorf("missing part")
+var errInvalidPart = fmt.Errorf("invalid part")
+
+// Authorize will run a batch of authorization evaluations.
+//
+// All the request parts (PARC) must be specified, but you can
+// specify [Variable] or [Ignore].  Varibles can be enumerated
```
kjamieson-sdm, line 87: nit: typo "Varibles" -> "Variables"

kjamieson-sdm (approved):
Looks good to me.

My review of the new "batch" interface was fairly superficial, given its experimental status and that it has already been more closely reviewed by others.

PR #35, opened 2024-09-16
Add the extension types, datetime and duration

apg:
See merged [RFC 80](https://github.com/cedar-policy/rfcs/pull/80).

This PR adds:
- The `datetime` and `duration` type specified in [RFC 80](https://github.com/cedar-policy/rfcs/pull/80)
- Associated methods for those types
- An implementation of operator overloading for `ComparableValue`s, such as Long, Datetime, Duration.

This does not extend ComparableValue to Decimal, as there's no RFC / specification to make that so (doing so would be straightforward, however).

Both `datetime` and `duration` are backed by an int64 and can be lifted from a Go `time.Time` and a `time.Duration` respectively. It would be possible to remove the struct wrapper around the int64, as no other state is stored. We aren't doing this to prevent accidental casting / type assertions that remove the semantic meaning of these values.

(Note: there was some previous discussion in the strongdm fork of this repo: https://github.com/strongdm/cedar-go/pull/3)

### DIFF
--- a/internal/consts/consts.go
+++ b/internal/consts/consts.go
@@ -7,6 +7,13 @@ const (
 	Context   = "context"
 )
 
+const (
+	MillisPerSecond = int64(1000)
+	MillisPerMinute = MillisPerSecond * 60
+	MillisPerHour   = MillisPerMinute * 60
+	MillisPerDay    = MillisPerHour * 24
+)
+
 func init() {
 	_ = 42
 }
--- a/internal/eval/comparable.go
+++ b/internal/eval/comparable.go
@@ -1,0 +1,17 @@
+package eval
+
+import "github.com/cedar-policy/cedar-go/types"
+
+// ComparableValue provides the interface that must be implemented to
+// support operator overloading of <, <=, >, and >=
+type ComparableValue interface {
+	types.Value
+
+	// LessThan returns true if the lhs is less than the rhs, and an
+	// error if the rhs is not comparable to the lhs
+	LessThan(types.Value) (bool, error)
+
+	// LessThan returns true if the lhs is less than or equal to the
+	// rhs, and an error if the rhs is not comparable to the lhs
+	LessThanOrEqual(types.Value) (bool, error)
+}
--- a/internal/eval/evalers.go
+++ b/internal/eval/evalers.go
@@ -22,7 +22,7 @@ func zeroValue() types.Value {
 type Env struct {
 	Entities                    types.Entities
 	Principal, Action, Resource types.Value
-	Context                     types.Value
+	Env                         types.Value
 
 	inCache map[inKey]bool
 }
@@ -74,6 +74,18 @@ func evalLong(n Evaler, env *Env) (types.Long, error) {
 	return l, nil
 }
 
+func evalLesser(n Evaler, env *Env) (types.Lesser, error) {
+	v, err := n.Eval(env)
+	if err != nil {
+		return nil, err
+	}
+	l, ok := v.(types.Lesser)
+	if !ok {
+		return nil, fmt.Errorf("%w: expected comparable value, got %v", ErrType, TypeName(v))
+	}
+	return l, nil
+}
+
 func evalString(n Evaler, env *Env) (types.String, error) {
 	v, err := n.Eval(env)
 	if err != nil {
@@ -108,6 +120,18 @@ func evalEntity(n Evaler, env *Env) (types.EntityUID, error) {
 		return types.EntityUID{}, err
 	}
 	return e, nil
+}
+
+func evalDatetime(n Evaler, env *Env) (types.Datetime, error) {
+	v, err := n.Eval(env)
+	if err != nil {
+		return types.Datetime{}, err
+	}
+	d, err := ValueToDatetime(v)
+	if err != nil {
+		return types.Datetime{}, err
+	}
+	return d, nil
 }
 
 func evalDecimal(n Evaler, env *Env) (types.Decimal, error) {
@@ -920,7 +944,7 @@ func (n *variableEval) Eval(env *Env) (types.Value, error) {
 	case consts.Resource:
 		return env.Resource, nil
 	default: // context
-		return env.Context, nil
+		return env.Env, nil
 	}
 }
 
@@ -1112,6 +1136,29 @@ func (n *decimalLiteralEval) Eval(env *Env) (types.Value, error) {
 	return d, nil
 }
 
+// datetimeLiteralEval
+type datetimeLiteralEval struct {
+	literal Evaler
+}
+
+func newDatetimeLiteralEval(literal Evaler) *datetimeLiteralEval {
+	return &datetimeLiteralEval{literal: literal}
+}
+
+func (n *datetimeLiteralEval) Eval(env *Env) (types.Value, error) {
+	literal, err := evalString(n.literal, env)
+	if err != nil {
+		return zeroValue(), err
+	}
+
+	d, err := types.ParseDatetime(string(literal))
+	if err != nil {
+		return zeroValue(), err
+	}
+
+	return d, nil
+}
+
 type ipLiteralEval struct {
 	literal Evaler
 }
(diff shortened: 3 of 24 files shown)


internal/eval/evalers.go
```diff
@@ -1194,10 +1276,14 @@ func newExtensionEval(name types.Path, args []Evaler) Evaler {
 			return newErrorEval(fmt.Errorf("%w: %s takes %d parameter(s)", errArity, name, i.Args))
 		}
 		switch {
-		case name == "ip":
-			return newIPLiteralEval(args[0])
+		case name == "datetime":
```
philhassey, line 1279: Optional: but it might be nice if all these strings were constants in the ExtMap package.  Something like ... 
```
const PathIP = "ip" 
```

Then the map could use those references as well.

internal/eval/evalers.go
```diff
@@ -74,6 +74,18 @@ func evalLong(n Evaler, env *Env) (types.Long, error) {
 	return l, nil
 }
 
+func evalComparableValue(n Evaler, env *Env) (types.ComparableValue, error) {
+	v, err := n.Eval(env)
+	if err != nil {
+		return nil, err
+	}
+	l, ok := v.(types.ComparableValue)
```
philhassey, line 82: Can you make ComparableValue be an interface that only exists within the evalers package?  I don't think the general public needs to be aware of it.

internal/eval/evalers.go
```diff
@@ -1218,7 +1304,285 @@ func newExtensionEval(name types.Path, args []Evaler) Evaler {
...
+func newToHoursEval(lhs Evaler) *toHoursEval {
+	return &toHoursEval{lhs: lhs}
+}
+
+func (n *toHoursEval) Eval(env *Env) (types.Value, error) {
+	lhs, err := evalDuration(n.lhs, env)
+	if err != nil {
+		return zeroValue(), err
+	}
+	return types.Long(lhs.Value / 1000 / 60 / 60), nil
```
philhassey, line 1529: Can all theses things end up as constants?  `lhs.Value / Hour` or something

types/datetime.go
```diff
@@ -0,0 +1,292 @@
...
+// "YYYY-MM-DDThh:mm:ssZ" (UTC)
+// "YYYY-MM-DDThh:mm:ss.SSSZ" (UTC with millisecond precision)
+// "YYYY-MM-DDThh:mm:ss(+/-)hhmm" (With timezone offset in hours and minutes)
+// "YYYY-MM-DDThh:mm:ss.SSS(+/-)hhmm" (With timezone offset in hours and minutes and millisecond precision)
+func ParseDatetime(s string) (Datetime, error) {
+	var (
+		year, month, day, hour, minute, second, milli int
+		i                                             int
+	)
+
```
philhassey, line 35: Max nitpick - maybe less linebreaks would feel a bit nicer.

types/datetime.go
```diff
@@ -0,0 +1,292 @@
...
+	// mm
+
+	if !(unicode.IsDigit(rune(s[i])) &&
+		unicode.IsDigit(rune(s[i+1]))) {
+		return Datetime{}, fmt.Errorf("%w: invalid minute", ErrDatetime)
+	}
+
+	minute = 10*int(rune(s[i])-'0') + int(rune(s[i+1])-'0')
+
+	i = i + 2
```
philhassey, line 136: Maybe it would be nice if there was a few private constants
`
const dateOffset = 3
const hourOffset = 6 
```
etc.

The changing of i, then using offsets from i is a bit hard to read.

File: types/datetime.go [hunk 6]
```diff
@@ -0,0 +1,292 @@
...
+	t = t.Add(sign * (offsetH + offsetM))
+	return Datetime{Value: t.UnixMilli()}, nil
+}
+
+func (a Datetime) Equal(bi Value) bool {
+	b, ok := bi.(Datetime)
+	return ok && a == b
+}
+
+func (a Datetime) Less(bi Value) bool {
```
philhassey, line 230: Less and LessEqual should return errors in case the types are not the same.

Also, I'd call them `LessThan` and `LessThanOrEqual` to better match the cedar operators list:
https://docs.cedarpolicy.com/policies/syntax-operators.html

types/datetime.go
```diff
@@ -0,0 +1,292 @@
...
+	if len(b) > 0 && b[0] == '"' {
+		if err := json.Unmarshal(b, &arg); err != nil {
+			return errors.Join(errJSONDecode, err)
+		}
+	} else {
+		// NOTE: cedar supports two other forms, for now we're only supporting the smallest implicit and explicit form.
+		// The following are not supported:
+		// "datetime(\"1970-01-01T00:00:00Z\")"
+		// {"fn":"datetime","arg":"1970-01-01T00:00:00Z"}
+		var res extValueJSON
```
philhassey, line 259: As long as this is in parity with what we did for Decimal/IPAddr it's okay for now.  But maybe add an issue somewhere about the deficit.

types/duration.go
```diff
@@ -0,0 +1,258 @@
...
+	"encoding/json"
+	"errors"
+	"fmt"
+	"math"
+	"strconv"
+	"time"
+	"unicode"
+)
+
+const (
```
philhassey, line 14: Ah, since I mentioned those constants elsewhere, maybe they can all live in the `internal/consts` package?

types/duration.go
```diff
@@ -0,0 +1,258 @@
...
+
+var unitOrder = []string{"d", "h", "m", "s", "ms"}
+
+// A Duration is a value representing a span of time in milliseconds.
+type Duration struct {
+	Value int64
+}
+
+// UnsafeDuration creates a duration via unsafe conversion from int, int64, float64
+func UnsafeDuration[T int | int64 | float64](v T) Duration {
```
philhassey, line 37: Maybe it should multiply by 1000 or something?  And the comment should indicate what it takes in, e.g. seconds?

philhassey: I'd probably just drop this function and only include FromStdDuration.

philhassey (commented):
Hope those help!

types/datetime.go
philhassey: Alternatively it could just take in the exact type only `Less(bi Datetime)`

Then you'd just do all the type checking work within the `evalers` package.

Could go either way on this.

types/datetime.go
apg: I'll just add it for duration/datetime, as they aren't hard to add.

types/datetime.go
apg: I think it's less desirable for the method to take a concrete type given there's an interface here. I'll do a type assertion in the implementation.

apg:
@philhassey I think all of your comments have been addressed.

internal/eval/comparable.go
```diff
@@ -0,0 +1,9 @@
+package eval
+
+import "github.com/cedar-policy/cedar-go/types"
+
+type ComparableValue interface {
+	types.Value
+	LessThan(types.Value) (bool, error)
+	LessThanEqual(types.Value) (bool, error)
```
philhassey, line 8: Can we change LessThanEqual to LessThanOrEqual, that better matches the cedar docs:

https://docs.cedarpolicy.com/policies/syntax-operators.html#function-lessThanOrEqual

types/duration.go
```diff
@@ -0,0 +1,264 @@
...
+
+	// We still have characters left, but no more units to assign.
+	if i < len(s) {
+		return Duration{}, fmt.Errorf("%w: invalid duration", ErrDuration)
+	}
+
+	return Duration{Value: negative * total}, nil
+}
+
+func (a Duration) Equal(bi Value) bool {
```
philhassey, line 118: All method and functions need a go-doc style docstring.

types/datetime.go
```diff
@@ -0,0 +1,281 @@
...
+	"time"
+	"unicode"
+)
+
+type Datetime struct {
+	// Value is a timestamp in milliseconds
+	Value int64
+}
+
+func UnsafeDatetime[T int | int64 | float64](v T) Datetime {
```
philhassey, line 18: I still think this function isn't useful.

types/value.go
```diff
@@ -4,8 +4,11 @@ import (
 	"fmt"
 )
 
+var ErrDatetime = fmt.Errorf("error parsing datetime value")
```
philhassey, line 7: I'm not sure if these need to be public, maybe they could be dropped in an internal package.

philhassey (commented):
Hope that helps!

types/datetime.go
apg: ha! I think I stopped using it and didn't delete it. I'll get rid of it.

philhassey:
I think this will be the next release, so maybe add in something to the README.md explaining the latest goodies in 0.3.2.  

And also add some explanation as to how datetime is an accepted RFC, but it's a bit bleeding edge (as discussed in the meeting yesterday)

types/datetime.go
apg: Apparently I didn't stop using it. 🤔 I _will_ stop using it.

types/datetime.go
```diff
@@ -0,0 +1,304 @@
...
+	}
+
+	t := time.Date(year, time.Month(month), day,
+		hour, minute, second,
+		int(time.Duration(milli)*time.Millisecond), time.UTC)
+	t = t.Add(offset)
+	return Datetime{Value: t.UnixMilli()}, nil
+}
+
+// Equal returns true if the lfs represents the same timestamp as the rhs.
```
philhassey, line 199: I think "lhs" instead of "lfs"

I feel like "lhs" and "rhs" is a bit odd.  Maybe have it phrased "Equal returns true if the input represents the same timestamp."

philhassey (approved):
Looks good, just a tiny text edit.  Can @patjakdev give this a quick review as well?

apg:
OK. At 100% coverage for all changes. There's one line that's not covered in cedar_unmarshal that is unrelated to my changes. I tried to figure out a quick way to reach it, but failed. :)

internal/eval/evalers.go
```diff
@@ -1218,7 +1305,299 @@ func newExtensionEval(name types.Path, args []Evaler) Evaler {
...
+func (n *comparableValueGreaterThanEval) Eval(env *Env) (types.Value, error) {
+	lhs, err := evalComparableValue(n.lhs, env)
+	if err != nil {
+		return zeroValue(), err
+	}
+	rhs, err := evalComparableValue(n.rhs, env)
+	if err != nil {
+		return zeroValue(), err
+	}
+	ok, err := lhs.LessThanOrEqual(rhs)
```
patjakdev, line 1386: tiny nit: the `ok` here maybe should be `isLessThanOrEqual`? `ok` implies something else to me.

internal/consts/consts.go
```diff
@@ -7,6 +7,13 @@ const (
 	Context   = "context"
 )
 
+const (
+	MillisPerDay    = int64(1000 * 60 * 60 * 24)
+	MillisPerHour   = int64(1000 * 60 * 60)
+	MillisPerMinute = int64(1000 * 60)
+	MillisPerSecond = int64(1000)
```
patjakdev, line 14: tiny nit: seems like these could be composed from each other

types/duration.go
```diff
@@ -0,0 +1,295 @@
...
+	"m":  consts.MillisPerMinute,
+	"s":  consts.MillisPerSecond,
+	"ms": 1,
+}
+
+var unitOrder = []string{"d", "h", "m", "s", "ms"}
+
+// A Duration is a value representing a span of time in milliseconds.
+type Duration struct {
+	Value int64
```
patjakdev, line 28: I think we should strongly consider making `Value` private. The accessor methods with units seem like the right public interface for this.

types/datetime.go
```diff
@@ -0,0 +1,302 @@
...
+	"fmt"
+	"strconv"
+	"time"
+	"unicode"
+)
+
+// Datetime represents a Cedar datetime value
+type Datetime struct {
+	// Value is a timestamp in milliseconds
+	Value int64
```
patjakdev, line 16: I think we should strongly consider keeping `Value` private and making an accessor method that gives you the time since the epoch in milliseconds.

patjakdev (approved):
Looks really good to me. Just a couple of nits.

types/datetime.go
apg: Changed. Note that Decimal still exposes this.

types/duration.go
apg: Changed. Note that Decimal still exposes this.

types/datetime.go
philhassey: I think Decimal will take a bit more effort, I've made an issue in the IDX project for us to handle that in a separate PR.

PR #38, opened 2024-09-18
Improve `Set.Contains()` performance by making `Set` a hash set-like data structure

patjakdev:
*Issue #, if available:* None

*Description of changes:*

This PR changes `types.Set` into a cheeseball open-addressable hash set by changing the underlying storage from a slice to a `map[uint64]types.Value` where the `uint64` key is a hash of the corresponding `types.Value`.

This required us to add a `hash()` function to the `Value()` interface. Non-trivial implementations of the `hash()` function use the `fnv` hash from the Go standard library, which is apparently the same one used (at least at one time) for Go's internal implementation of `map`.

In order to make this safe, we also had to make `Record` and `Set` immutable, which was generally a good quality of life improvement in terms of cognitive overhead when thinking about how to deal with `Value`s and allowed us to remove the `deepClone()` method of the `Value` interface. It does, however, mean that in order to construct a `Set` or `Record`, one now needs to go through the `NewSet()` or `NewRecord()` constructor function, respectively.

Collisions are resolved by simply adding 1 to the computed hash of the `Value` until a free slot is found in the map. This works trivially, so long as the `Set` remains immutable because to no values can be removed from the map and thus no holes in the open addressing chain can be created.


.github/workflows/build_and_test.yml
```diff
@@ -22,7 +22,7 @@ jobs:
         run: go build -v ./...
 
       - name: Test
-        run: go test -v ./...
```
patjakdev, line 25: Unrelated, but the output of the workflow has become completely unwieldy.

internal/eval/evalers_test.go
```diff
@@ -1234,6 +1234,28 @@ func TestContainsAnyNode(t *testing.T) {
 			})
 		}
 	}
+
+	t.Run("not quadratic", func(t *testing.T) {
```
patjakdev, line 1238: This test demonstrates that a worst-case search for the intersection between two sets was previously quadratic.

x/exp/batch/batch.go
```diff
@@ -220,10 +220,11 @@ func doBatch(ctx context.Context, be *batchEvaler) error {
 	// then loop the current variable
 	loopEnv := *be.env
 	u := be.Variables[0]
-	_, chPrincipal := cloneSub(be.env.Principal, u.Key, nil)
-	_, chAction := cloneSub(be.env.Action, u.Key, nil)
-	_, chResource := cloneSub(be.env.Resource, u.Key, nil)
-	_, chContext := cloneSub(be.env.Context, u.Key, nil)
+	dummyVal := types.True
+	_, chPrincipal := cloneSub(be.env.Principal, u.Key, dummyVal)
```
patjakdev, line 224: @philhassey, I think the use of this dummy value here is okay. Let me know if you have any other more clever ideas how to get this to work. The problem is that `Set` tries to call `.hash()` on `nil` here, which obviously doesn't work.

patjakdev:
I'm considering using the `maphash` [package](https://pkg.go.dev/hash/maphash@go1.23.1) for the non-trivial hashing instead of `fnv`. It seems like it's fit for purpose.

types/set.go
```diff
@@ -2,19 +2,94 @@ package types
...
+			}
+			hashKey++
+		}
+	}
+
+	// Special case hashVal for empty set to 0 so that the return value of Value.hash() of Set{} and NewSet([]Value{})
+	// are the same
+	var hashVal uint64
+	if len(set) > 0 {
+		slices.Sort(hashes)
```
apg, line 51: Seems like we could probably just sum all the element's `hash()` and be faster without much additional fear of collisions? Basically, the hash() of other values becomes more important.

Note: I looked for reasons to believe this is true and found that Java implements `hashCode()` for `AbstractSet` this way:

> Returns the hash code value for this set. The hash code of a set is defined to be the sum of the hash codes of the elements in the set, where the hash code of a null element is defined to be zero. This ensures that s1.equals(s2) implies that s1.hashCode()==s2.hashCode() for any two sets s1 and s2, as required by the general contract of [Object.hashCode()](https://docs.oracle.com/javase/8/docs/api/java/lang/Object.html#hashCode--).

patjakdev: Good idea!

types/set.go
```diff
@@ -3,18 +3,83 @@ package types
...
+
+// NewSet takes a slice of Values and stores a clone of the values internally.
+func NewSet(v []Value) Set {
+	orig := make([]Value, len(v))
+	set := make(map[uint64]Value, len(v))
+
+	for i, vv := range v {
+		orig[i] = vv
+
+		hash := vv.hash()
```
patjakdev, line 25: Another thought. Could we just call `vv.String()` here and hash the string output of the `Value`? We'd have to check if the `String()` output is always the same for any equivalent instantiation of `Value`...

Just a thought before I step away for the day...

apg:
We could. Seems cheaper to sum `hash()` rather than build a string and hash that though. Summing `hash()` seems generic enough to work for all non-atomic types.On Sep 18, 2024, at 17:02, Patrick Jakubowski ***@***.***> wrote:﻿
@patjakdev commented on this pull request.



In types/set.go:
> -type Set []Value
+// A Set is an immutable collection of elements that can be of the same or different types.
+type Set struct {
+	s       map[uint64]Value
+	hashVal uint64
+}
+
+// NewSet takes a slice of Values and stores a clone of the values internally.
+func NewSet(v []Value) Set {
+	orig := make([]Value, len(v))
+	set := make(map[uint64]Value, len(v))
+
+	for i, vv := range v {
+		orig[i] = vv
+
+		hash := vv.hash()

Another thought. Could we just call vv.String() here and hash the string output of the Value? We'd have to check if the String() output is always the same for any equivalent instantiation of Value...
Just a thought before I step away for the day...

x/exp/batch/batch.go
philhassey: That's fine.

internal/eval/evalers_test.go
```diff
@@ -1244,16 +1266,16 @@ func TestRecordLiteralNode(t *testing.T) {
 		result types.Value
 		err    error
 	}{
-		{"empty", map[types.String]Evaler{}, types.Record{}, nil},
+		{"empty", map[types.String]Evaler{}, types.NewRecord(types.RecordMap{}), nil},
```
philhassey, line 1269: types.Record{} could work

types/decimal_internal_test.go
```diff
@@ -0,0 +1,21 @@
...
+	t.Parallel()
+	t.Run("hash", func(t *testing.T) {
+		t.Parallel()
+
+		testutil.Equals(t, UnsafeDecimal(42).hash(), UnsafeDecimal(42).hash())
+		testutil.Equals(t, UnsafeDecimal(-42).hash(), UnsafeDecimal(-42).hash())
+
+		// This isn't necessarily true for all values of Decimal, but we want to ensure we aren't just returning the
+		// same hash value for Decimal.hash() for every instance.
+		testutil.Equals(t, false, UnsafeDecimal(42).hash() == UnsafeDecimal(1337).hash())
```
philhassey, line 19: FatalIf would be better

types/record.go
```diff
@@ -2,25 +2,73 @@ package types
...
+func (r Record) Get(s String) (Value, bool) {
+	v, ok := r.m[s]
+	return v, ok
+}
 
 // Equals returns true if the records are Equal.
 func (a Record) Equal(bi Value) bool {
 	b, ok := bi.(Record)
-	if !ok || len(a) != len(b) {
+	if !ok || len(a.m) != len(b.m) {
```
philhassey, line 67: Add a.m.hash != b.m.hash

types/record_test.go
```diff
@@ -57,18 +57,126 @@ func TestRecord(t *testing.T) {
...
+		m := types.RecordMap{"foo": types.Long(42), "bar": types.Long(1337)}
+		r := types.NewRecord(m)
+
+		delete(m, "foo")
+		m["bar"] = types.Long(42)
+
+		testutil.Equals(t, r.Len(), 2)
+
+		got := types.RecordMap{}
+		r.Iterate(func(k types.String, v types.Value) bool {
```
philhassey, line 175: maybe add `r.Map() -> RecordMap`

maybe for sets `r.Slice() -> []Value`

types/set.go
```diff
@@ -3,18 +3,83 @@ package types
...
-type Set []Value
+// A Set is an immutable collection of elements that can be of the same or different types.
+type Set struct {
+	s       map[uint64]Value
+	hashVal uint64
+}
+
+// NewSet takes a slice of Values and stores a clone of the values internally.
+func NewSet(v []Value) Set {
+	orig := make([]Value, len(v))
```
philhassey, line 19: org is unused

types/set.go
philhassey: I'd vote against this, since creating the full String is so costly.

types/set.go
```diff
@@ -3,18 +3,83 @@ package types
...
+	set := make(map[uint64]Value, len(v))
+
+	for i, vv := range v {
+		orig[i] = vv
+
+		hash := vv.hash()
+		hashKey := hash
```
philhassey, line 26: also drop hashKey

types/set.go
```diff
@@ -23,13 +88,13 @@ func (as Set) Equal(bi Value) bool {
 	if !ok {
 		return false
 	}
-	for _, a := range as {
-		if !bs.Contains(a) {
-			return false
-		}
+
+	if len(as.s) != len(bs.s) {
```
philhassey, line 92: add in hash comparison to the first if

types/set_internal_test.go
```diff
@@ -0,0 +1,130 @@
...
+		})
+
+		// These tests don't necessarily hold for all values of Set, but we want to ensure we are considering
+		// different aspects of the Set, which these particular tests demonstrate.
+
+		t.Run("extra element", func(t *testing.T) {
+			t.Parallel()
+			s1 := NewSet([]Value{Long(42), Long(1337)})
+			s2 := NewSet([]Value{Long(42), Long(1337), Long(1)})
+			testutil.Equals(t, false, s1.hash() == s2.hash())
```
philhassey, line 98: ErrorIf

x/exp/batch/batch.go
```diff
@@ -328,29 +329,55 @@ func cloneSub(r types.Value, k types.String, v types.Value) (types.Value, bool)
...
-			if vv, delta := cloneSub(vv, k, v); delta {
-				if !cloned {
-					t = maps.Clone(t) // intentional shallow clone
-					cloned = true
-				}
-				t[kk] = vv
+		hasDeltas := false
+
+		// First pass, check for deltas
+		t.Iterate(func(_ types.String, vv types.Value) bool {
```
philhassey, line 335: with .Map() added this could be a single Iterate call with some tweaking

philhassey (approved):
LGTM, some minor tweaks.

PR #39, opened 2024-09-19
cedar: project public types, constructors, and constants from the types package through the cedar package for convenience

patjakdev:
*Issue #, if available:* None

*Description of changes:* For convenience of use, this PR adds projections of all the public types, constructors, and constants from the `types` package through the `cedar` package.


types.go
```diff
@@ -0,0 +1,80 @@
...
+const (
+	True  = types.True
+	False = types.False
+)
+
+const (
+	DecimalPrecision = types.DecimalPrecision
+)
+
+// Constructors
```
philhassey, line 42: Check how this reads in pkgsite.  It might be nice to either copy the docs, or provide a [link] to the docs for each of the functions in the types package.

patjakdev: Good call. I've fleshed out the docs for the rest of the functions.

types/entities.go
```diff
@@ -21,7 +21,10 @@ type Entity struct {
 }
 
 func (e Entities) MarshalJSON() ([]byte, error) {
-	s := e.toSlice()
+	s := maps.Values(e)
```
patjakdev, line 24: Unrelated change. I just noticed that these could be inlined while I was looking for functions to export...

philhassey (approved):
LGTM

PR #40, opened 2024-09-20
Create MapSet[T] and ImmutableMapSet[T] types and change Entity.Parents to be a ImmutableMapSet[EntityUID]

patjakdev:
*Issue #, if available:* None

*Description of changes:*

This PR adds a generic `MapSet[T]` and `ImmutableMapSet[T]` for all the places we want the semantics of a set but are annoyed by the inconvenience of using the `map[T]struct{}` idiom in Go.

In particular, the parents of an `Entity` have set semantics. With the current slice-based implementation, time complexity of the evaluation of `in` is linear. Utilizing a `ImmutableMapSet[EntityUID]` instead gives us constant time complexity on average, which is better. That being said, for small sets, that constant time is actually slightly greater than the time it takes to perform a linear search. I will consider addressing that in a future PR, but for now I'm happy to just have set semantics for `Entity` parents.

The `ImmutableMapSet[EntityUID]` is projected publicly through the `types` package as `EntityUIDSet`, with appropriate constructors also projected. For now, at least, I thought it prudent not to allow external users to import and use `MapSet[T]` directly.

I've also replaced any existing internal usages of `map[T]struct{}` in the module with `MapSet[T]` or `ImmutableMapSet[T]` as appropriate.

Important notes:
1. `MapSet[T]` has no ordering guarantees. This applies to `MarshalJSON` as well, meaning there's no predictable JSON encoding for a given `Set`. Instead of using `comparable`, we could invent an `Item` interface that embeds `comparable` and also has a `Less()` method, but then this set couldn't easily be used for native types like `string`, which feels like a pretty lame drawback.
2. While `MapSet` is implemented in `internal` and therefore can't be directly imported by external users, the `Entity.Parents` is now of type `MapSet[EntityUID]` and therefore changes to the `MapSet[T]` interface should be considered breaking changes. I'm open to feedback on a better way to structure this. I considered putting `MapSet` in `types`, but thought that might be confusing...especially because I have some desire to extract an `ImmutableHashSet` out of `types.Set` in the future.


types/entities.go
```diff
@@ -13,13 +13,6 @@ import (
 // the Entity (it must be the same as the UID within the Entity itself.)
 type Entities map[EntityUID]*Entity
 
-// An Entity defines the parents and attributes for an EntityUID.
```
patjakdev, line 16: I extracted this because I always have trouble remembering where the heck this struct is defined.

authorize_test.go
```diff
@@ -1,68 +1,68 @@
-package cedar
+package cedar_test
```
patjakdev, line 1: This is good hygiene. None of the tests in this file need to interact with any private bits in the package.

authorize_test.go
```diff
@@ -71,435 +71,435 @@ func TestIsAuthorized(t *testing.T) {
...
 		{
 			Name:   "permit-requires-entities-parent-success",
 			Policy: `permit(principal,action,resource) when { principal in parent::"bob" };`,
-			Entities: Entities{
-				cuzco: &Entity{
+			Entities: cedar.Entities{
+				cuzco: &cedar.Entity{
 					UID:     cuzco,
-					Parents: []EntityUID{types.NewEntityUID("parent", "bob")},
+					Parents: cedar.NewEntityUIDSetFromSlice([]cedar.EntityUID{cedar.NewEntityUID("parent", "bob")}),
```
patjakdev, line 142: Here's an example of the breaking change.

I'll admit that it's a bit of a mouthful...

authorize_test.go
```diff
@@ -71,435 +71,435 @@ func TestIsAuthorized(t *testing.T) {
...
 		{
 			Name:   "permit-principal-in",
 			Policy: `permit(principal in team::"osiris",action,resource);`,
-			Entities: Entities{
-				cuzco: &Entity{
+			Entities: cedar.Entities{
+				cuzco: &cedar.Entity{
 					UID:     cuzco,
-					Parents: []EntityUID{types.NewEntityUID("team", "osiris")},
+					Parents: cedar.NewEntityUIDSetFromSlice([]cedar.EntityUID{cedar.NewEntityUID("team", "osiris")}),
```
patjakdev, line 169: Another example.

authorize_test.go
```diff
@@ -71,435 +71,435 @@ func TestIsAuthorized(t *testing.T) {
...
 		{
 			Name:   "permit-action-in",
 			Policy: `permit(principal,action in scary::"stuff",resource);`,
-			Entities: Entities{
-				dropTable: &Entity{
+			Entities: cedar.Entities{
+				dropTable: &cedar.Entity{
 					UID:     dropTable,
-					Parents: []EntityUID{types.NewEntityUID("scary", "stuff")},
+					Parents: cedar.NewEntityUIDSetFromSlice([]cedar.EntityUID{cedar.NewEntityUID("scary", "stuff")}),
```
patjakdev, line 196: A couple more examples here and on line 212.

authorize_test.go
```diff
@@ -600,183 +600,183 @@ func TestIsAuthorized(t *testing.T) {
...
 			Name:   "when-is-in",
 			Policy: `permit(principal,action,resource) when { resource is Resource in Parent::"id" };`,
-			Entities: Entities{
-				NewEntityUID("Resource", "table"): &Entity{
-					UID:     NewEntityUID("Resource", "table"),
-					Parents: []EntityUID{types.NewEntityUID("Parent", "id")},
+			Entities: cedar.Entities{
+				cedar.NewEntityUID("Resource", "table"): &cedar.Entity{
+					UID:     cedar.NewEntityUID("Resource", "table"),
+					Parents: cedar.NewEntityUIDSetFromSlice([]cedar.EntityUID{cedar.NewEntityUID("Parent", "id")}),
```
patjakdev, line 757: The last example in this file

authorize_test.go
philhassey: Maybe receive a variadic, to cut out the `[]cedar.EntityUID{` ?  Though TBH, may not matter too much since it's going to be usually handled via JSON marshalling.

Maybe the words `FromSlice` aren't really needed.

internal/eval/evalers.go
```diff
@@ -987,31 +983,28 @@ func entityInOne(env *Env, entity types.EntityUID, parent types.EntityUID) bool
...
+				if !ok || p.Parents.Len() == 0 || k == entity || known.Contains(k) {
+					return true
 				}
 				todo = append(todo, k)
-				if known == nil {
-					known = map[types.EntityUID]struct{}{}
-				}
-				known[k] = struct{}{}
-			}
+				known.Add(k)
```
philhassey, line 1005: I like the automatic deferred allocation.

philhassey: Makes this whole bit of code more readable?

internal/eval/evalers_test.go
```diff
@@ -1860,10 +1861,10 @@ func TestEntityIn(t *testing.T) {
 
 		entityMap := types.Entities{}
 		for i := 0; i < 100; i++ {
-			p := []types.EntityUID{
+			p := sets.NewMapSetFromSlice([]types.EntityUID{
```
philhassey, line 1864: Maybe if the package was called `mapset` you could just do `mapset.New(variadic...)`

internal/sets/mapset.go
```diff
@@ -0,0 +1,153 @@
...
+
+// MapSet is a struct that adds some convenience to the otherwise cumbersome map[T]struct{} idiom used in Go to
+// implement sets of comparable types.
+type MapSet[T comparable] struct {
+	m map[T]struct{}
+}
+
+// NewMapSet returns a MapSet ready for use. Optionally, a desired size for the MapSet can be passed as an argument,
+// as in the argument to make() for a map type.
+func NewMapSet[T comparable](args ...int) MapSet[T] {
```
philhassey, line 22: Make this could be `MakeMapSet` since it is kind of like `make(map[],n)`

internal/sets/mapset.go
```diff
@@ -0,0 +1,153 @@
...
+	var m map[T]struct{}
+	if size > 0 {
+		m = make(map[T]struct{}, size)
+	}
+
+	return MapSet[T]{m: m}
+}
+
+// NewMapSetFromSlice creates a MapSet of size len(items) and calls AddSlice(items) on it.
+func NewMapSetFromSlice[T comparable](items []T) MapSet[T] {
```
philhassey, line 41: And maybe this could just be `New(variadic...)`

internal/sets/mapset.go
philhassey: or just `Make` if we call the package `mapset`

internal/sets/mapset.go
```diff
@@ -0,0 +1,153 @@
...
+}
+
+// Add an item to the set. Returns true if the item did not exist in the set.
+func (h *MapSet[T]) Add(item T) bool {
+	if h.m == nil {
+		h.m = map[T]struct{}{}
+	}
+
+	_, exists := h.m[item]
+	h.m[item] = peppercorn
```
philhassey, line 54: Maybe do an `if` around the exists and just return true.  Then return false in the other case.

internal/sets/mapset.go
```diff
@@ -0,0 +1,153 @@
...
+	return !exists
+}
+
+// AddSlice adds a slice of items to the set, returning true if any new items were added to the set.
+func (h *MapSet[T]) AddSlice(items []T) bool {
```
philhassey, line 59: Call this `Append` maybe?

philhassey: And be variadic?

internal/sets/mapset.go
```diff
@@ -0,0 +1,153 @@
...
+	modified := false
+	for _, i := range items {
+		modified = h.Add(i) || modified
+	}
+	return modified
+}
+
+// Remove an item from the Set. Returns true if the item existed in the set.
+func (h *MapSet[T]) Remove(item T) bool {
```
philhassey, line 68: Could just have the one Remove and have it be variadic?

internal/sets/mapset.go
```diff
@@ -0,0 +1,153 @@
...
+}
+
+// Contains returns whether the item exists in the set
+func (h MapSet[T]) Contains(item T) bool {
+	_, exists := h.m[item]
+	return exists
+}
+
+// Intersection returns the items common to both h and o.
+func (h MapSet[T]) Intersection(o MapSet[T]) MapSet[T] {
```
philhassey, line 90: This is interesting, when do we use it?

philhassey: The one place we use it, we just care about the Len.  So maybe instead of paying the cost of allocation, have this function instead be

`Intersects(o) bool` then it can quit early and not allocate.

types.go
```diff
@@ -85,6 +86,17 @@ func NewEntityUID(typ EntityType, id String) EntityUID {
 	return types.NewEntityUID(typ, id)
 }
 
+// NewEntityUIDSet returns an EntityUIDSet ready for use. Optionally, a desired size for the set can be passed as an
```
philhassey, line 89: Not sure how useful this function is.

philhassey: Seems it is really only used in tests.  I'd just stick with the second, make it variadic.  Get rid of `FromSlice`

types/entity_uid.go
```diff
@@ -91,3 +93,16 @@ func (i ImplicitlyMarshaledEntityUID) MarshalJSON() ([]byte, error) {
...
+// NewEntityUIDSet returns an EntityUIDSet ready for use. Optionally, a desired size for the EntityUIDSet can be passed
+// as an argument, as in the argument to make() for a map type.
+func NewEntityUIDSet(args ...int) EntityUIDSet {
+	return sets.NewMapSet[EntityUID](args...)
+}
+
+// NewEntityUIDSetFromSlice creates a EntityUIDSet of size len(items) and calls AddSlice(items) on it.
+func NewEntityUIDSetFromSlice(items []EntityUID) EntityUIDSet {
+	return sets.NewMapSetFromSlice[EntityUID](items)
+}
```
philhassey, line 108: Same thoughts on the naming, variadic stuff.

internal/parser/cedar_unmarshal.go
```diff
@@ -139,11 +140,11 @@ func (p *parser) annotations() (ast.Annotations, error) {
 
 }
 
-func (p *parser) annotation(a *ast.Annotations, known map[string]struct{}) error {
+func (p *parser) annotation(a *ast.Annotations, known *sets.MapSet[string]) error {
```
philhassey, line 143: Probably doesn't need to be a pointer since it contains the map.

philhassey: Though, I suppose if the map was nil, it would need to be a pointer.

x/exp/batch/batch.go
```diff
@@ -375,11 +382,11 @@ func cloneSub(r types.Value, k types.String, v types.Value) (types.Value, bool)
 	return r, false
 }
 
-func findVariables(found map[types.String]struct{}, r types.Value) {
+func findVariables(found *sets.MapSet[types.String], r types.Value) {
```
philhassey, line 385: I wonder if sets.NewMapSet()... should return a pointer since the Add method leans on the ability to be able to modify things?

philhassey: Annoyingly though that costs an allocation in `in` lookups.  So I guess it's a funny tradeoff.  Hmm.

philhassey (approved):
A lot of little nitpicks here, glad to discuss.  Great improvement to a lot of stuff!

internal/sets/mapset.go
patjakdev: Groovy, if you feel like that's a better name I'm all for it.

philhassey:
> Set has no ordering. This applies to MarshalJSON as well, meaning there's no predictable JSON encoding for a given Set. Instead of using comparable, we could invent an Item interface that embeds comparable and also has a Less() method, but then this set couldn't easily be used for native types like string, which feels like a pretty lame drawback.

I think having a consistent JSON marshaller is important because having things change randomly is really annoying and can cause unnecessary updates happen when a diff is detected by something somewhere.

In the case of JSON marshaling, I don't think performance is the key metric.  So I think what you could do is get a Marshal of the bytes of all the values.  Then sort those slices of bytes, then emit the "[a,b,c]" in that order.

>While MapSet is implemented in internal and therefore can't be directly imported by external users, the Entity.Parents is now of type MapSet[EntityUID] and therefore changes to the MapSet[T] interface should be considered breaking changes. I'm open to feedback on a better way to structure this. I considered putting MapSet in types, but thought that might be confusing...especially because I have some desire to extract an ImmutableHashSet out of types.Set in the future.

I thought the Entity.Parents was going to be immutable?

I agree that having `type X internal.Something` isn't great, but I think we can defer fixing that for a while.  We have similar problems in our public `ast` package.

internal/eval/evalers.go
```diff
@@ -973,11 +974,6 @@ func newInEval(lhs, rhs Evaler) Evaler {
 	return &inEval{lhs: lhs, rhs: rhs}
 }
 
-func hasKnown(known map[types.EntityUID]struct{}, k types.EntityUID) bool {
-	_, ok := known[k]
-	return ok
-}
-
 func entityInOne(env *Env, entity types.EntityUID, parent types.EntityUID) bool {
 	key := inKey{a: entity, b: parent}
```
philhassey, line 978: I thought the benchmarks indicated the cache wasn't providing much value?

Fine to leave changing the cache to another PR though.

internal/eval/evalers.go
```diff
@@ -1020,31 +1013,27 @@ func entityInOneWork(env *Env, entity types.EntityUID, parent types.EntityUID) b
...
 		if fe, ok := env.Entities[candidate]; ok {
-			for _, k := range fe.Parents {
-				if _, ok := parents[k]; ok {
-					return true
-				}
+			if fe.Parents.Intersects(parents) {
+				return true
 			}
-			for _, k := range fe.Parents {
+			fe.Parents.Iterate(func(k types.EntityUID) bool {
```
philhassey, line 1028: This is fine, but as a point of interest, I wonder if using a closure causes an allocation.

internal/mapset/immutable_test.go
```diff
@@ -0,0 +1,166 @@
...
+		testutil.OK(t, err)
+		testutil.Equals(t, string(out), "[]")
+	})
+
+	t.Run("encode json", func(t *testing.T) {
+		s := Immutable(1, 2, 3)
+
+		out, err := json.Marshal(s)
+
+		correctOutputs := []string{
```
philhassey, line 124: See other comment, let's not allow this.

internal/mapset/immutable_test.go
```diff
@@ -0,0 +1,166 @@
...
+	})
+
+	t.Run("two elements", func(t *testing.T) {
+		s := Immutable[int](1, 2)
+		testutil.Equals(t, s.Contains(1), true)
+		testutil.Equals(t, s.Contains(2), true)
+		testutil.Equals(t, s.Contains(3), false)
+	})
+
+	t.Run("deduplicate elements", func(t *testing.T) {
```
philhassey, line 34: If there is a .Len() method this might be a better place to prove deduplication.

internal/mapset/mapset.go
```diff
@@ -0,0 +1,141 @@
...
+}
+
+// Add an item to the set. Returns true if the item did not exist in the set.
+func (h *MapSet[T]) Add(item T) bool {
+	if h.m == nil {
+		h.m = map[T]struct{}{}
+	}
+
+	_, exists := h.m[item]
+	h.m[item] = peppercorn
```
philhassey, line 57: Can you `if _,exists := { return }`

internal/mapset/mapset.go
```diff
@@ -0,0 +1,141 @@
...
+	for item := range h.m {
+		if !o.Contains(item) {
+			return false
+		}
+	}
+	return true
+}
+
+// MarshalJSON serializes a MapSet as a JSON array. Order is non-deterministic.
+func (h MapSet[T]) MarshalJSON() ([]byte, error) {
```
philhassey, line 125: Also update the comment when it becomes deterministic.

internal/mapset/mapset_test.go
```diff
@@ -0,0 +1,234 @@
...
+
+func TestHashSet(t *testing.T) {
+	t.Run("empty set contains nothing", func(t *testing.T) {
+		s := Make[int]()
+		testutil.Equals(t, s.Len(), 0)
+		hashSetMustNotContain(t, s, 1)
+
+		s = Make[int](10)
+		testutil.Equals(t, s.Len(), 0)
+		hashSetMustNotContain(t, s, 1)
```
philhassey, line 23: Maybe checking for contains `0` would be more interesting, since it's the zero value.

internal/mapset/mapset_test.go
```diff
@@ -0,0 +1,234 @@
...
+
+		s2.Add(4)
+		testutil.Equals(t, s1.Equal(s2), false)
+
+		s2.Remove(3)
+		testutil.Equals(t, s1.Equal(s2), false)
+
+		s1.Add(4)
+		s1.Remove(3)
+		testutil.Equals(t, s1.Equal(s2), true)
```
philhassey, line 93: Maye drop in a `s2.Equal(s1)`

policy.go
```diff
@@ -97,7 +97,7 @@ func (p *Policy) Position() Position {
 	return Position(p.ast.Position)
 }
 
-// SetFilename sets the filename of this policy.
+// SetFilename mapset the filename of this policy.
```
philhassey, line 100: `s/sets/mapset` too hard

types/entity.go
```diff
@@ -0,0 +1,31 @@
...
+	UID        EntityUID    `json:"uid"`
+	Parents    EntityUIDSet `json:"parents"`
+	Attributes Record       `json:"attrs"`
+}
+
+// MarshalJSON serializes Entity as a JSON object, using the implicit form of EntityUID encoding to match the Rust
+// SDK's behavior.
+func (e Entity) MarshalJSON() ([]byte, error) {
+	parents := make([]ImplicitlyMarshaledEntityUID, 0, e.Parents.Len())
+	e.Parents.Iterate(func(p EntityUID) bool {
```
philhassey, line 16: Throw a slices.Sort on the parents slice to keep this consistent.

philhassey: Maybe add a test to prove the point.

types/set.go
```diff
@@ -89,7 +89,7 @@ func (s Set) Slice() []Value {
 	return maps.Values(s.s)
 }
 
-// Equal returns true if the sets are Equal.
+// Equal returns true if the mapset are Equal.
```
philhassey, line 92: s/sets/mapset too hard

internal/eval/evalers.go
patjakdev: Yeah, I was going to do that in a separate PR.

philhassey (approved):
Looks good, a few minor tweaks would be nice.

PR #41, opened 2024-09-23
Invert the special casing of JSON marshaling for Value

patjakdev:
*Issue #, if available:* None

*Description of changes:*

Entities and extension functions have both an "implict" and "explict" form of JSON serialization, where the "implicit" form is only valid if the serialized data can be unambiguously determined to be of a given type. This can happen positionally (e.g. EntityUIDs in the serialization of the Entities object) or via knowledge gleaned from a schema (which we don't yet support).

It's _always_ safe to encode an entity or extension function via the "explicit" form, so let's make that the default. This also alleviates the types that don't have implicit and explicit forms from having to implement `ExplicitMarshalJSON()`.

With this approach, callers who know that it's safe to use the implicit form can do so and everyone else will get the safe explicit form. The two existing callers that use the implicit form are the JSON serialization of Entities and the scope operators which always take EntityUIDs. I've updated these to use the implicit form, which matches the behavior of the Rust SDK.

At this time, I don't really see any reason to ever encode extension functions via the implicit form. Even if we had schema support, it just seems to save a few bytes and makes the JSON encoding mildly more legible, which shouldn't really be a goal of the JSON. FWIW, the Rust SDK doesn't seem to ever encode extension functions via the implicit form.

### DIFF
--- a/internal/json/json.go
+++ b/internal/json/json.go
@@ -1,6 +1,8 @@
 package json
 
 import (
+	"encoding/json"
+
 	"github.com/cedar-policy/cedar-go/types"
 )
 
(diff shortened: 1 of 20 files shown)


philhassey (approved):
LGTM

PR #43, opened 2024-09-30
internal/mapset: relax the constraint on the argument to MapSet.Equals to just be a Container instance

patjakdev:
*Issue #, if available:* None

*Description of changes:* This makes it easier to compare sets of different types.


apg (approved):
Given that we don't support infinite streams of entities by design, this seems safe to me. :)

PR #45, opened 2024-10-01
internal/eval: remove the inCache

patjakdev:
*Issue #, if available:* None

*Description of changes:*

We don't currently have a benchmark that's telling us that we need this cache. In fact, several of our benchmarks show that for large, shallow entity graphs, the inCache actually slows down authorizations and batch evaluations because of all the extra allocation that goes on to build the map.

For deep entity graphs, such a cache might be useful, but I think instead that we'll try to put such a cache in the entity graph itself by keeping track of the transitive closure of every entity's parents. That way, the cache will be effective across multiple authorizations.

### DIFF
--- a/authorize.go
+++ b/authorize.go
@@ -19,13 +19,13 @@ const (
 // IsAuthorized uses the combination of the PolicySet and Entities to determine
 // if the given Request to determine Decision and Diagnostic.
 func (p PolicySet) IsAuthorized(entityMap Entities, req Request) (Decision, Diagnostic) {
-	c := eval.InitEnv(&eval.Env{
+	env := eval.Env{
 		Entities:  entityMap,
 		Principal: req.Principal,
 		Action:    req.Action,
 		Resource:  req.Resource,
 		Context:   req.Context,
-	})
+	}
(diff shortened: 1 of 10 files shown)


authorize.go
```diff
@@ -35,7 +35,7 @@ func (p PolicySet) IsAuthorized(entityMap Entities, req Request) (Decision, Diag
 	// - For permit, all permits must be run to collect annotations
 	// - For forbid, forbids must be run to collect annotations
 	for id, po := range p.policies {
-		result, err := po.eval.Eval(c)
+		result, err := po.eval.Eval(env)
```
apg, line 38: I question not passing Env by reference here, mostly because Env will get large, and that's likely to result in a lot of copying? I don't believe that Go has an optimization that can detect potentially large copies and silently convert to pass by reference, but I could be wrong. If so, then, this is fine.

apg: I forgot that Env is "Just" a tuple of pointers. This shouldn't be a big deal then. Nevermind.

apg (approved):

PR #49, opened 2024-11-06
Create a safer interface for the Decimal type`

patjakdev:
*Issue #, if available:* None

*Description of changes:*

This PR adds safe constructors for `Decimal`, allowing callers to pass an integer, a floating point value, or a decimal value as expressed via a significant and exponent.

Overflow and underflow are checked to ensure that invalid values can't be passed. However, care must be taken when using `NewDecimalFromFloat()` as values with a high number of significant digits can suffer from precision loss.

A `Cmp()` method has also been added to `Decimal`, which is required in order to allow for the evaluation of the decimal comparison operators.

*Breaking changes:*
* The internal representation of Decimal as an `int64` has been made private to allow future flexibility in implementation.
* The `UnsafeDecimal()` constructor has been removed. Callers should use the new, safe constructors instead.
* The DecimalPrecision constant has been made private


types/decimal.go
```diff
@@ -1,28 +1,91 @@
...
+// Decimal above those sizes, use the NewDecimal constructor.
+func NewDecimalFromFloat[T constraints.Float](f T) (Decimal, error) {
+	f = f * decimalPrecision
+	if f > math.MaxInt64 {
+		return Decimal{}, fmt.Errorf("%w: value %v would overflow", ErrDecimal, f)
+	} else if f < math.MinInt64 {
+		return Decimal{}, fmt.Errorf("%w: value %v would underflow", ErrDecimal, f)
+	}
+
+	return NewDecimal(int64(f), -4)
```
apg, line 78: You've already munged the number to the correct int64. Doesn't seem necessary to do the work again by calling `NewDecimal` here?

types.go
```diff
@@ -109,8 +106,14 @@ func NewSet(s []types.Value) Set {
 	return types.NewSet(s)
 }
 
-// UnsafeDecimal creates a decimal via unsafe conversion from int, int64, float64.
-// Precision may be lost and overflows may occur.
-func UnsafeDecimal[T int | int64 | float64](v T) Decimal {
-	return types.UnsafeDecimal(v)
+func NewDecimal(i int64, exponent int) (Decimal, error) {
```
philhassey, line 109: Can you add docs for all these functions?

philhassey: Also, I see the other ParseIP, etc funcs aren't wrapped here, maybe those could be done as well?

types/decimal.go
```diff
@@ -1,28 +1,91 @@
...
 	"strconv"
 	"unicode"
+
+	"golang.org/x/exp/constraints"
 )
 
+// decimalPrecision is the precision of a Decimal.
+const decimalPrecision = 10000
+
+var DecimalMax = Decimal{value: math.MaxInt64}
```
philhassey, line 18: Do these really need to be public?

types/decimal.go
```diff
@@ -1,28 +1,91 @@
...
+
+	return NewDecimal(int64(f), -4)
+}
+
+// Cmp returns
+//
+//	-1 if d is less than other,
+//	 0 if d equals other,
+//	+1 if d is greater than other.
+func (d Decimal) Cmp(other Decimal) int {
```
philhassey, line 86: Can we rename this to `Compare` to match the go library's function name?

types/decimal.go
apg: or at least `const`?

types/decimal_test.go
```diff
@@ -105,23 +105,183 @@ func TestDecimal(t *testing.T) {
...
+			{-92233720368548, 1},
+			{-922337203685477581, 2},
+			{-10, 14},
+			{-1, 15},
+		}
+		for _, tt := range tests {
+			t.Run(fmt.Sprintf("%ve%v", tt.in, tt.exp), func(t *testing.T) {
+				t.Parallel()
+				_, err := types.NewDecimal(tt.in, tt.exp)
+				testutil.ErrorIs(t, err, types.ErrDecimal)
```
philhassey, line 233: Does ErrDecimal need to be public?  Or is it adequate for it just to emit errors?

types/decimal.go
patjakdev: Nice. I'll make the change. I generally like to err on the side of going through the "one true" constructor for types, but it does seem kind of silly in this case.

types.go
patjakdev: Good catch. I wish the story for projecting functions from another package was a little bit neater on the Go documentation side.

types/decimal.go
patjakdev: Yes. Not sure what I was thinking here.

types/decimal.go
philhassey: The tests can just assert that they got an error back and not be particular about the type of error.  Or the errors could live in an internal package that the tests can access as well as the package can access.

I suppose the harm is everything we make public must be there for all eternity as of 1.0, so unless there is an argument for making it public, it's better to keep it private.

philhassey: Oh sorry, I was talking about the public errors, but the same gist applies here WRT internal package.

types/decimal_test.go
patjakdev: Let's take care of that in another PR that hides all the errors.

types/decimal.go
```diff
@@ -1,138 +1,150 @@
...
-	fraction := int64(0)
-	fractionDigits := 0
-	for ; i < len(s); i++ {
-		c = rune(s[i])
-		if !unicode.IsDigit(c) {
-			return Decimal{}, fmt.Errorf("%w: unexpected character %s", ErrDecimal, strconv.QuoteRune(c))
+// ParseDecimal takes a string representation of a decimal number and converts it into a Decimal type.
+func ParseDecimal(s string) (Decimal, error) {
+	decimalIndex := strings.Index(s, ".")
+	if decimalIndex < 0 {
```
philhassey, line 96: Are we sure the "." is required?

patjakdev: Good question. The Cedar Rust implementation [does seem to require it](https://github.com/cedar-policy/cedar/blob/dc9163c6f71d72633495f80262ab2d0b266138f7/cedar-policy-core/src/extensions/decimal.rs#L59).

philhassey (approved):
LGTM

PR #51, opened 2024-11-07
Unify naming conventions across map/set container types

patjakdev:
*Issue #, if available:* None

*Description of changes:*

* Add() for upserting items into the container; returns bool indicating if the key existed prior to the call
* Get() for retrieving items
* Delete() for removing items; returns bool indicating if the key existed prior to the call

I also made a few miscellaneous changes like renaming files, fixing typos, and adjusting the README to not use the types package any more.


philhassey (approved):
I think s/Delete/Remove would be better.

This PR should also cover the `EntityLoader` interface and rename it to `EntityGetter`

patjakdev:
> I think s/Delete/Remove would be better.
> 
> This PR should also cover the `EntityLoader` interface and rename it to `EntityGetter`

Sounds good, I've made the `Delete()` -> `Remove()` change. We'll tackle the `EntityLoader` in another PR.

PR #53, opened 2024-11-07
types: change Set() to take variadic arguments to match EntityUIDSet()

patjakdev:
*Issue #, if available:*

*Description of changes:*

In addition to making the two functions have a similar shape, this also cuts down quite a bit on the annoying `[]types.Value{}` wrapper detritus that we have to put around literal arguments.


philhassey (approved):
LGTM

PR #57, opened 2024-11-08
ast: Add AST builder methods for datetime-related types and operators

patjakdev:
*Issue #, if available:* None

*Description of changes:* Add missing AST builder methods for datetime and duration literals as well as their associated methods.

### DIFF
--- a/ast/value.go
+++ b/ast/value.go
@@ -2,6 +2,7 @@ package ast
 
 import (
 	"net/netip"
+	"time"
 
 	"github.com/cedar-policy/cedar-go/internal/ast"
 	"github.com/cedar-policy/cedar-go/types"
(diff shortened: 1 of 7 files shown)


ast/value.go
```diff
@@ -79,6 +80,14 @@ func IPAddr[T netip.Prefix | types.IPAddr](i T) Node {
 	return wrapNode(ast.IPAddr(types.IPAddr(i)))
 }
 
+func Datetime(t time.Time) Node {
+	return Value(types.FromStdTime(t))
+}
+
+func Duration(d time.Duration) Node {
+	return Value(types.FromStdDuration(d))
+}
```
patjakdev, line 89: It would be nice if these could take their counterparts from the `types` library like the other builder functions, but alas they cannot because the `types` types don't have the same underlying structure as the stdlib types.

I looked into making `types.Datetime` a wrapper around `time.Time` so that we could support both types as an argument here, but you end up giving `types.Datetime` at least one of the same footguns that `time.Time` has: the `==` operator does not return true if two instances represent the same moment of time in two different locations. Instead, you need to use the `.Equal()` method to resolve that.

apg (approved):

PR #59, opened 2024-11-08
Expose extension function calls in AST package

### LINKED ISSUE #54: Allow extension call nodes to be programmatically added to the AST
Today, users can add extension nodes (`decimal`, `ip`, `datetime`, and `duration`) to the graph, but only if they have a valid instance of the Go value type associated with those nodes. For example:

```
duration1 := types.DurationFromMillis(100)
duration2 := types.DurationFromMillis(1000)
policy := ast.Permit().When(ast.Value(duration1).LessThan(ast.Value(duration2)))
```

However, users might also want to just take a string, plunk it into the AST unparsed, and let the evaluator be the one that handles any errors in parsing just like if a Cedar policy had an invalid string specified in Cedar text.

This issue proposes adding some new public AST functions to add such unevaluated extension method call nodes to the AST. Their use would look something like this:

```
policy := ast.Permit().When(ast.DurationExtensionCall("100").LessThan(ast.DurationExtensionCall("1000"))
```

Under the hood, these helpers would just call the existing `internal.ast.ExtensionCall()` function.

micahhausler:
*Issue #, if available:*

Resolves #54

*Description of changes:*

This change adds support for specifically calling extension functions in the AST package.

### DIFF
--- a/ast/ast_test.go
+++ b/ast/ast_test.go
@@ -11,10 +11,8 @@ import (
 	"github.com/cedar-policy/cedar-go/types"
 )
 
-// These tests serve mostly as examples of how to translate from Cedar text into programmatic AST construction. They
-// don't verify anything.
-func TestAstExamples(t *testing.T) {
-	t.Parallel()
+// This example shows how you can construct polcies with the ast package
+func Example() {
 
 	johnny := types.NewEntityUID("User", "johnny")
 	sow := types.NewEntityUID("Action", "sow")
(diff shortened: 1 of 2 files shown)


ast/ast_test.go
```diff
@@ -10,10 +10,8 @@ import (
 	"github.com/cedar-policy/cedar-go/types"
 )
 
-// These tests serve mostly as examples of how to translate from Cedar text into programmatic AST construction. They
-// don't verify anything.
-func TestAstExamples(t *testing.T) {
-	t.Parallel()
+// This example shows how you can construct polcies with the ast package
+func ExampleNode() {
```
patjakdev, line 14: Yes, thank you! We've been meaning to do this.

I think we could actually elevate these up to just `Example()` so that they appear right at the top of the documentation.

patjakdev (approved):
Love it! Thank you @micahhausler.

micahhausler:
Fixed conflicts, updated test example to package-level `Example()`.

PR #63, opened 2024-11-12
ast: restructure ast package documentation

patjakdev:
*Issue #, if available:* None

*Description of changes:*

Along with moving the example tests out to example_test.go, I've reorganized them a bit and added some output checking that ensures that our documentation remains correct if we change the underlying marshaling to Cedar text. It also looks real purdy in the pkgsite documentation.

This necessitated projecting Cedar marshaling for ast.Policy. I decided to bring along JSON marshaling as well for consistency.


philhassey (approved):
LGTM

PR #64, opened 2024-11-12
types: clean up JSON unmarshaling story for extension types

patjakdev:
*Issue #, if available:* None

*Description of changes:*

Create a generic `unmarshalExtensionValue()` helper function that supports all three forms of extension value JSON encoding:

* Explicit: `{ "__extn": { "fn": "decimal", "arg": "1234.5678" } }`
* Implicit object: `{ "fn": "decimal", "arg": "1234.5678" } }`
* Implicit string: `"1234.5678"`

Use it to flesh out the implementations of `UnmarshalJSON` for `IPAddr` and `Decimal`. Note that the implicit decoding will never actually be invoked until we have schema-driven decoding for entities.

### DIFF
--- a/types/json.go
+++ b/types/json.go
@@ -142,7 +142,12 @@ func unmarshalExtensionValue[T any](b []byte, extName string, parse func(string)
 		if res.Extn == nil {
 			// If we didn't find an Extn, maybe it's just an extn.
 			var res2 extn
+
+			// N.B. we already know that this JSON unmarshals successfully as a struct, so it's
+			// not actually possible to get an error here. Hence, it's safe to ignore the return
+			// value.
 			_ = json.Unmarshal(b, &res2)
+
 			// We've tried Ext.Fn and Fn, so no good.
 			if res2.Fn == "" {
 				return zeroT, errJSONExtNotFound
--- a/types/json.go
+++ b/types/json.go
@@ -3,6 +3,7 @@ package types
 import (
 	"bytes"
 	"encoding/json"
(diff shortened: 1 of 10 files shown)


types/json.go
```diff
@@ -125,3 +126,42 @@ func UnmarshalJSON(b []byte, v *Value) error {
 	}
 	return nil
 }
+
+func unmarshalExtensionValue[T any](b []byte, extName string, parse func(string) (T, error)) (T, error) {
+	var zeroT T
+	var arg string
```
apg, line 132: nit: you could combine this into one var declaration

```suggestion
	var (
		zeroT T
		arg string
	)
```

patjakdev: I think I prefer the two line version...

types/json.go
```diff
@@ -125,3 +126,42 @@ func UnmarshalJSON(b []byte, v *Value) error {
...
+		}
+	} else {
+		var res extValueJSON
+		if err := json.Unmarshal(b, &res); err != nil {
+			return zeroT, errors.Join(errJSONDecode, err)
+		}
+		if res.Extn == nil {
+			// If we didn't find an Extn, maybe it's just an extn.
+			var res2 extn
+			_ = json.Unmarshal(b, &res2)
```
patjakdev, line 145: Note to self: document why it's okay to ignore this error...

philhassey: Or not!

philhassey (approved):
LGTM

PR #66, opened 2024-11-13
Sprinkle more high-level documentation around

patjakdev:
*Issue #, if available:* None

*Description of changes:*

Add high level package documentation to a number of packages and an `example_test.go` to the main cedar package which will show up in the Go documentation.


apg:
> I'd love if we could embed the example_test.go output directly into README.md so we don't have to duplicate the code, but I can't find an easy way to do that.

PR #68, opened 2024-11-19
Use comparable evaluators in partial evaluation and folding

kjamieson-sdm:
This PR changes partial evaluation and folding to use the comparable value evaluators in place of the long comparison evaluators. This fixes issues where partial evaluation and folding would produce an error with less than/greater than/less than or equal/greater than or equal comparisons were performed against comparable values other than long (such as datetime).


philhassey (approved):
LGTM

PR #69, opened 2024-11-21
Allow to use types defined from bool, string, int and int64 as ast Values

forgems:
Addresses #IDX-461

This change will simplify usage of values that underlying type is based on bool, string, int and int64 as Node.

Currently if we have a value of the custom type ex `type CustomBool bool` we have to convert it to bool or types.Boolean first. By adding ~ at type constraints to Boolean, String and Long functions we can use variables of derived type directly without the need of conversion


ast/ast_test.go
```diff
@@ -11,6 +11,11 @@ import (
 	"github.com/cedar-policy/cedar-go/types"
 )
 
+type CustomString string
```
philhassey, line 14: Can these types be put inside the `TestASTByTable` function?  Other than that, LGTM.

philhassey (approved):

ast/ast_test.go
forgems: Thanks. Moved

PR #71, opened 2024-11-29
IDX-470: Change PolicySet.IsAuthorized to take an EntityGetter interface in place of an EntityMap

kjamieson-sdm:
This changes the `PolicySet` `IsAuthorized()` method to take an `EntityGetter` interface instead of a concrete `EntityMap` instance, allowing alternative entity getter implementations to be passed to the authorization.

### DIFF
--- a/README.md
+++ b/README.md
@@ -137,8 +137,9 @@ Generated documentation for the latest version of the Go implementation can be a
 If you're looking to integrate Cedar into a production system, please be sure the read the [security best practices](https://docs.cedarpolicy.com/other/security.html)
 
 ## Backward Compatibility Considerations
-
-x/exp - code in this directory is not subject to the semantic version constraints of the rest of the module and breaking changes may be made at any time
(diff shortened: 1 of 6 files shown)


kjamieson-sdm:
While very unlikely (IMO), this could technically be a breaking change if some existing caller is passing a `map[EntityUID]Entity` to `IsAuthorized` instead of a `EntityMap`.

If we think that risk is too great, we can fall back to the alternative of introducing some new `IsAuthorized` or `Authorizer` implementation under x/exp instead.

philhassey:
My only thought is maybe if we add something to the README just noting our definition of non-breaking is that would be nice.  Something like:

- we may add variadics to functions that don't have them in cases where we want to expand the arguments of a function / method
- we may change concrete types to interfaces in cases where we want to expand the variety of arguments a function / method can take

philhassey (approved):
LGTM, with that one add to README

PR #72, opened 2024-12-03
Improve batch error handling

philhassey:
This PR changes two things:

- Makes the error handling for context errors more consistent, it will return an error if you cancel in the last batch request.
- Adds an error to the batch callback signature.  Any error will cause the batch processing to cease and the error to be returned.

### DIFF
--- a/x/exp/batch/batch_test.go
+++ b/x/exp/batch/batch_test.go
@@ -382,6 +382,33 @@ func TestBatchErrors(t *testing.T) {
 		testutil.Equals(t, total, 1)
 	})
 
+	t.Run("lastContextCancelled", func(t *testing.T) {
+		ctx, cancel := context.WithCancel(context.Background())
+		var total int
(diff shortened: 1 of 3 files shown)


x/exp/batch/batch_test.go
```diff
@@ -356,22 +358,77 @@ func TestBatchErrors(t *testing.T) {
...
+			total++
+			if total == 2 {
+				return errWant
+			}
+			return nil
+		},
+		)
+		testutil.ErrorIs(t, err, errWant)
+		testutil.Equals(t, total, 2)
+	})
```
kjamieson-sdm, line 423: Maybe add one more test case here that both returns an error and does a context cancel, and verifies that the returned error is both? As I think no tests here are currently exercising the behaviour that the callback error is joined with the context error.

kjamieson-sdm (approved):
Looks good, with perhaps the addition of the one more test case suggested.

PR #76, opened 2024-12-05
types: add go native conversion methods to several types

philhassey:
Datetime, Duration, and Decimal lacked convenient conversion methods.  This PR adds those methods.


patjakdev (approved):
Looks great!

types/datetime.go
```diff
@@ -274,6 +274,11 @@ func (a Datetime) Milliseconds() int64 {
 	return a.value
 }
 
+// Time returns the time.Time representation of a Datetime.
+func (a Datetime) Time() time.Time {
+	return time.UnixMilli(a.value)
```
patjakdev, line 279: Hmmmm, check on the Location for this `time.Time`. From a cursory reading of the code, I think it might be in the local timezone, which would be wrong.

philhassey: Enforcing UTC now, thanks!

PR #77, opened 2024-12-06
RFC-82: Implement support for Entity Tags

### LINKED ISSUE #47: Add support for RFC 82 (entity tags)
[RFC 82](https://github.com/cedar-policy/rfcs/blob/emina/entity-tags/text/0082-entity-tags.md) was recently accepted and is supported in the Rust library as of version [4.2.0](https://github.com/cedar-policy/cedar/releases/tag/v4.2.0).

### LINKED ISSUE #52: Upstream Integration Test Corpus Modified
The upstream integration test corpus at https://raw.githubusercontent.com/cedar-policy/cedar-integration-tests/main/corpus-tests.tar.gz has been updated. Please integrate the changes into the local copy.

adowair:
This pull request will implement support for Cedar Entity Tags per [RFC-82](https://github.com/cedar-policy/rfcs/blob/7fcdbb509dd2709e03897fe40accbefd8fdb56f3/text/0082-entity-tags.md).

This pull request will also update the [corpus tests](https://github.com/cedar-policy/cedar-integration-tests) to the latest version: https://github.com/cedar-policy/cedar-integration-tests/pull/12

Closes #47
Closes #52

#### Test Coverage
```console
$ git checkout main
$ go test -count 1 -cover ./... | grep -v -E 'coverage: 100|no test files'
ok      github.com/cedar-policy/cedar-go/ast    0.257scoverage: 98.1% of statements
$ git checkout tag-support
$ go test -count 1 -cover ./... | grep -v -E 'coverage: 100|no test files'
ok      github.com/cedar-policy/cedar-go/ast    1.315scoverage: 98.2% of statements
```


internal/eval/fold_test.go
```diff
@@ -473,6 +473,26 @@ func TestFoldPolicy(t *testing.T) {
...
+		},
+		{
+			"opGetTagEntity",
+			ast.Permit().When(ast.EntityUID("T", "1").GetTag(ast.String("key"))),
+			ast.Permit().When(ast.EntityUID("T", "1").GetTag(ast.String("key"))),
+		},
+		{
+			"opHasTagEntity",
+			ast.Permit().When(ast.EntityUID("T", "1").HasTag(ast.String("key"))),
+			ast.Permit().When(ast.False()),
```
adowair, line 494: @philhassey @patjakdev notice that `hasTag()` nodes are _folded_ when there are no entities in the context. This is because `<non_existing_medallion>.hasTag("someTag")` evaluates to `false`.

authorize.go
```diff
@@ -19,6 +19,9 @@ const (
 // IsAuthorized uses the combination of the PolicySet and Entities to determine
 // if the given Request to determine Decision and Diagnostic.
 func (p PolicySet) IsAuthorized(entities types.EntityGetter, req Request) (Decision, Diagnostic) {
+	if entities == nil {
+		entities = types.EntityMap{}
+	}
```
patjakdev, line 24: Oh Golang...

authorize_test.go
```diff
@@ -784,7 +795,7 @@ func TestIsAuthorized(t *testing.T) {
 	for _, tt := range tests {
 		tt := tt
 		t.Run(tt.Name, func(t *testing.T) {
-			t.Parallel()
+			// t.Parallel()
```
patjakdev, line 798: Please add this back.

authorize_test.go
```diff
@@ -33,6 +34,16 @@ func TestIsAuthorized(t *testing.T) {
...
+		{
+			Name:      "nil-entity-map",
+			Policy:    `permit(principal,action,resource);`,
+			Principal: cuzco,
+			Action:    dropTable,
+			Resource:  cedar.NewEntityUID("table", "whatever"),
+			Context:   cedar.Record{},
+			Want:      true,
+			DiagErr:   0,
+		},
```
patjakdev, line 46: Can you explicitly initialize `Entities` to `nil` here?

Also, I guess I would name the test `nil-entity-getter`.

corpus_test.go
```diff
@@ -244,11 +245,27 @@ func TestCorpusRelated(t *testing.T) {
...
+			resource
+			) when {
+			true && (resource.hasTag("A"))
+			};`,
+			types.EntityMap{cedar.NewEntityUID("a", ""): cedar.Entity{Attributes: cedar.NewRecord(cedar.RecordMap{"A": types.False})}},
+			cedar.Request{Principal: cedar.NewEntityUID("a", ""), Action: cedar.NewEntityUID("Action", "action"), Resource: cedar.NewEntityUID("a", "'")},
+			cedar.Deny,
+			nil,
+			nil,
+		},
```
patjakdev, line 268: I'm not 100% convinced we should check this in. I think it would make sense instead to add a pared down version of this to `TestIsAuthorized()`.

For example, I don't think we actually care about the duplicate `Action::"action"` in the action set. You want to validate some very specific behaviors about `getTag()` and `hasTag()`.

internal/eval/evalers.go
```diff
@@ -804,6 +805,73 @@ func (n *hasEval) Eval(env Env) (types.Value, error) {
...
+func (n *getTagEval) Eval(env Env) (types.Value, error) {
+	eid, err := evalEntity(n.lhs, env)
+	if err != nil {
+		return zeroValue(), err
+	}
+
+	var unspecified types.EntityUID
+	if eid == unspecified {
+		return zeroValue(), fmt.Errorf("cannot access tag `%s` of %w", n.rhs, errUnspecifiedEntity)
+	}
```
patjakdev, line 826: I think I would call this a "zero" ID rather than an "unspecified" ID.

That being said, it appears that entity type names cannot be empty, but the id part of the string can be. The Cedar text unmarshaler wouldn't allow this entity to exist, but I think the JSON unmarshaler and `NewEntityUID()` would.

I think we should actually file a bug on the JSON unmarshaler for allowing empty type names. We should also really enforce the type name being non-empty in `NewEntityUID()`, but adding an error return type to that function would be a breaking change. Maybe we could panic?

All that is to say, I'm not sure how important it really is to consider this case in the evaler.

internal/eval/evalers.go
```diff
@@ -1139,9 +1207,13 @@ func newExtensionEval(name types.Path, args []Evaler) Evaler {
...
 	if i, ok := extensions.ExtMap[name]; ok {
 		if i.Args != len(args) {
-			return newErrorEval(fmt.Errorf("%w: %s takes %d parameter(s)", errArity, name, i.Args))
+			return newErrorEval(fmt.Errorf("%w: %s takes %d parameter(s), but %d provided", errArity, name, i.Args, len(args)))
 		}
 		switch {
+		case name == "hasTag":
+			return newHasTagEval(args[0], args[1])
+		case name == "getTag":
+			return newGetTagEval(args[0], args[1])
```
patjakdev, line 1216: `hasTag` and `getTag` are not extension methods. I would think they should be implemented as first class operators, like `contains()` and its ilk.

internal/eval/evalers_test.go
```diff
@@ -1448,6 +1448,175 @@ func TestHasNode(t *testing.T) {
...
+			newLiteralEval(unknownTag),
+			zeroValue(),
+			errTagAccess},
+		{"UnknownEntity",
+			newLiteralEval(types.NewEntityUID("unknownType", unknownID)),
+			newLiteralEval(knownTag),
+			zeroValue(),
+			errEntityNotExist},
+		{"UnspecifiedEntity",
+			newLiteralEval(types.NewEntityUID("", "")),
```
patjakdev, line 1509: Yeah, this totally shouldn't be allowed.

internal/eval/fold_test.go
patjakdev: Isn't this...wrong? Maybe I'm missing something, but we can't be allowed to fold this to false because during folding we don't know which entities actually exist.

The `has` operator works like I expect. Even though it returns false if the entity doesn't exist, it doesn't fold away here...(see `opHasEntity` above).

internal/extensions/extensions.go
```diff
@@ -8,6 +8,9 @@ type extInfo struct {
 }
 
 var ExtMap = map[types.Path]extInfo{
+	"hasTag": {Args: 2, IsMethod: true},
+	"getTag": {Args: 2, IsMethod: true},
+
```
patjakdev, line 13: As I mentioned above, these methods should be treated as  first-class operators, not extension methods.

types/entity.go
```diff
@@ -11,6 +11,7 @@ type Entity struct {
 	UID        EntityUID    `json:"uid"`
 	Parents    EntityUIDSet `json:"parents"`
 	Attributes Record       `json:"attrs"`
+	Tags       Record       `json:"tags,omitempty"`
```
patjakdev, line 14: You determined that `omitempty` is useless here, right?

types/entity_tag_test.go
```diff
@@ -0,0 +1,46 @@
...
+import (
+	"testing"
+	"time"
+
+	"github.com/cedar-policy/cedar-go"
+	"github.com/cedar-policy/cedar-go/internal/testutil"
+	"github.com/cedar-policy/cedar-go/types"
+)
+
+func TestEntityTagMarshalJSON(t *testing.T) {
```
patjakdev, line 12: I would put this in `types.entity_test.go`. I don't see a reason for it to live in its own file, really.

patjakdev (changes requested):

authorize.go
philhassey: Lol, I think ... slightly better might be (again, this is Go):

```
var zero types.EntityMap
entities = zero
```

Which is because (and feel free to Go playground / check on this explanation) when you just say `var zero types.EntityMap` you are yielding what amounts to a typed nil pointer.  However, when you do `entities = types.EntityMap{}` you are allocating an empty map.

In our case, probably either would work fine, but the `var zero` more precisely replicates the original behavior and avoids an unnecessary allocation.

internal/extensions/extensions.go
adowair: Thank you, done here: https://github.com/cedar-policy/cedar-go/pull/77/commits/4f433e34c428f2de144c83035ba1fd0a313a904c

internal/eval/evalers.go
adowair: Yes, done here: https://github.com/cedar-policy/cedar-go/pull/77/commits/4f433e34c428f2de144c83035ba1fd0a313a904c.

types/entity_tag_test.go
adowair: Yes, done here: https://github.com/cedar-policy/cedar-go/pull/77/commits/9074d5b00f44be4346c13ac8b7270a09b06c541a.

types/entity.go
adowair: Yes, removed here: https://github.com/cedar-policy/cedar-go/pull/77/commits/9074d5b00f44be4346c13ac8b7270a09b06c541a.

internal/eval/evalers_test.go
adowair: I've renamed this to `ZeroEntity`—while it's not a possible call path, we do need this test to maintain 100% net new code coverage. Done here: https://github.com/cedar-policy/cedar-go/pull/77/commits/9074d5b00f44be4346c13ac8b7270a09b06c541a.

internal/eval/evalers.go
adowair: Yes, I think that makes sense, thanks for expounding on it!

I've renamed this to `zero` here: https://github.com/cedar-policy/cedar-go/pull/77/commits/9074d5b00f44be4346c13ac8b7270a09b06c541a.

corpus_test.go
adowair: Yes, done here: https://github.com/cedar-policy/cedar-go/pull/77/commits/9074d5b00f44be4346c13ac8b7270a09b06c541a.

authorize_test.go
adowair: Yes, done here: https://github.com/cedar-policy/cedar-go/pull/77/commits/9074d5b00f44be4346c13ac8b7270a09b06c541a.

authorize_test.go
adowair: Yes, done here: https://github.com/cedar-policy/cedar-go/pull/77/commits/9074d5b00f44be4346c13ac8b7270a09b06c541a.

ast/ast_test.go
```diff
@@ -347,6 +347,16 @@ func TestASTByTable(t *testing.T) {
 			ast.Permit().When(ast.Long(42).Has("key")),
 			internalast.Permit().When(internalast.Long(42).Has("key")),
 		},
+		{
+			"opGetTag",
+			ast.Permit().When(ast.Long(42).GetTag(ast.String("key"))),
```
patjakdev, line 352: nit: I think we should strive for these to be "valid" ASTs, so maybe change those `Long`s to `EntityUID`s.

authorize.go
patjakdev: Agreed with Phil on avoiding the extra allocation, as nitpicky as it may be.

I'm not sure who would ever pass a nil `EntityGetter`.

internal/eval/fold_test.go
```diff
@@ -473,6 +473,26 @@ func TestFoldPolicy(t *testing.T) {
...
+		{
+			"opHasTag",
+			ast.Permit().When(ast.Long(42).HasTag(ast.String("key"))),
+			ast.Permit().When(ast.Long(42).HasTag(ast.String("key"))),
+		},
+		{
+			"opGetTagEntity",
+			ast.Permit().When(ast.EntityUID("T", "1").GetTag(ast.String("key"))),
+			ast.Permit().When(ast.EntityUID("T", "1").GetTag(ast.String("key"))),
+		},
```
patjakdev, line 490: It feels like this ought to be the `opGetTag` test and the current `opGetTag` test should be `opGetTagInvalidType` test or the like.

internal/eval/partial_test.go
```diff
@@ -1258,6 +1282,63 @@ func TestPartialBasic(t *testing.T) {
 	}
 }
 
+func TestPartialWithContext(t *testing.T) {
+	t.Parallel()
+	tests := []struct {
+		name       string
+		context    types.Value
+		entitities types.EntityGetter
```
patjakdev, line 1290: ahem

internal/parser/cedar_marshal.go
```diff
@@ -364,6 +380,8 @@ func astNodeToMarshalNode(astNode ast.IsNode) IsNode {
 		return NodeTypeNot{v, UnaryNode{}}
 	case ast.NodeTypeAccess:
 		return NodeTypeAccess{v}
+	case ast.NodeTypeGetTag:
+		return NodeTypeGetTag{v}
```
patjakdev, line 384: Did the two cases get separated by accident? It doesn't seem to be alphabetical or some other order I can discern.

x/exp/ast/node.go
```diff
@@ -67,6 +67,10 @@ type NodeTypeHas struct {
 	StrOpNode
 }
 
+type NodeTypeHasTag struct {
```
patjakdev, line 70: Keep it one one line?

x/exp/batch/batch.go
```diff
@@ -155,6 +155,9 @@ func Authorize(ctx context.Context, ps *cedar.PolicySet, entities types.EntityGe
 	case request.Context == nil:
 		return fmt.Errorf("%w: context", errMissingPart)
 	}
+	if entities == nil {
```
patjakdev, line 158: Same comment about avoiding allocation

patjakdev (approved):
Nice!!! Just a couple little nitpicks.

internal/eval/partial_test.go
adowair: oh my goodness

internal/eval/fold.go
```diff
@@ -109,6 +109,32 @@ func fold(n ast.IsNode) ast.IsNode {
 				return ast.NodeTypeHas{StrOpNode: ast.StrOpNode{Arg: nodes[0], Value: v.Value}}
 			},
 		)
+	case ast.NodeTypeGetTag:
+		return tryFold(
+			[]ast.IsNode{v.Left, v.Right},
+			func(values []types.Value) Evaler {
+				if _, ok := values[0].(types.EntityUID); ok {
```
philhassey, line 116: You could probably just return an error always for both this and HasTag

```
func(values []types.Value) Evaler {
    return newErrorEval(fmt.Errorf("fold.GetTag.EntityUID"))
}
```

types/entity.go
```diff
@@ -33,10 +34,12 @@ func (e Entity) MarshalJSON() ([]byte, error) {
 		UID        ImplicitlyMarshaledEntityUID   `json:"uid"`
 		Parents    []ImplicitlyMarshaledEntityUID `json:"parents"`
 		Attributes Record                         `json:"attrs"`
+		Tags       Record                         `json:"tags"`
```
philhassey, line 37: @patjakdev did you mention that for Rust parity this should be `omitempty` ?

philhassey: I guess, either way, a test showing it does what we want it to do for the empty tags case would be nice.

philhassey: (I see a test exists for this, so that's nice.)

philhassey (approved):
LGTM - just a few tiny nits left

philhassey:
There is a lint error to address as well:

image

internal/eval/fold_test.go
adowair: Yes, correct on all accounts! Fixed this up here: https://github.com/cedar-policy/cedar-go/pull/77/commits/4f433e34c428f2de144c83035ba1fd0a313a904c.

types/entity.go
adowair: Yes, we deliberated on this here: https://github.com/strongdm/cedar-go/pull/11#discussion_r1870037794

internal/parser/cedar_marshal.go
adowair: I've separated them to each respectively be right after the equivalent attribute operation.

internal/eval/fold.go
adowair: Yes, done here: https://github.com/cedar-policy/cedar-go/pull/77/commits/36fd3e7573bac8103b345a0c8d413d6eef37dccb.

x/exp/batch/batch.go
adowair: Yes, done here: https://github.com/cedar-policy/cedar-go/pull/77/commits/36fd3e7573bac8103b345a0c8d413d6eef37dccb.

x/exp/ast/node.go
adowair: Yes, done here: https://github.com/cedar-policy/cedar-go/pull/77/commits/36fd3e7573bac8103b345a0c8d413d6eef37dccb.

internal/eval/partial_test.go
adowair: Fixed here: https://github.com/cedar-policy/cedar-go/pull/77/commits/36fd3e7573bac8103b345a0c8d413d6eef37dccb.

internal/eval/fold_test.go
adowair: Yes, done here: https://github.com/cedar-policy/cedar-go/pull/77/commits/36fd3e7573bac8103b345a0c8d413d6eef37dccb.

ast/ast_test.go
adowair: Yes, done here: https://github.com/cedar-policy/cedar-go/pull/77/commits/36fd3e7573bac8103b345a0c8d413d6eef37dccb.

adowair:
Notice that this is a slightly breaking change w.r.t. the backward compatibility of JSON marshalling. Now `tags: {}` will be emitted as a part of entity struct, since empty structs cannot be omitted by the `omitempty` tag. Go `v1.24` will introduce an `omitzero` tag which solves this exact problem.

Issue #2, opened 2024-03-08
Schema Support


tpaulus:
cedar-go does not have parity, as noted in the README, with cedar-rust with regard to loading and including the schema in authorization requests. It would be nice to see cedar-go have support for this, as it enables additional use cases, such as the one below.

Use Case:
In order to express Action Hierarchy, where a Role contains one or more actions, a cedar schema needs to be loaded and used in the evaluation of policies. For example, the following schema allows policies to permit the `SuperAdmin` action, and have the PARC reference `BillingAdmin` and the outcome be Allow.

```json
{
  "Role": {
    "entityTypes": {},
    "actions": {
      "SuperAdmin": {
        "appliesTo": {
          "principalTypes": [],
          "resourceTypes": []
        }
      },
      "MinimalAccountAccess": {
        "memberOf": [
          {
            "id": "SuperAdmin"
          }
        ],
        "appliesTo": {
          "principalTypes": [],
          "resourceTypes": []
        }
      },
      "Admin": {
        "memberOf": [
          {
            "id": "SuperAdmin"
          }
        ],
        "appliesTo": {
          "principalTypes": [],
          "resourceTypes": []
        }
      },
      "AdminReadOnly": {
        "memberOf": [
          {
            "id": "Admin"
          }
        ],
        "appliesTo": {
          "principalTypes": [],
          "resourceTypes": []
        }
      },
      "BillingAdmin": {
        "memberOf": [
          {
            "id": "Admin"
          }
        ],
        "appliesTo": {
          "principalTypes": [],
          "resourceTypes": []
        }
      },
      "BillingAdminReadOnly": {
        "memberOf": [
          {
            "id": "BillingAdmin"
          },
          {
            "id": "AdminReadOnly"
          }
        ],
        "appliesTo": {
          "principalTypes": [],
          "resourceTypes": []
        }
      }
    }
  }
}
```

jmccarthy:
Hi @tpaulus - we'll leave this issue open and update it once we know more about the timeline for Schema support

PR #80, opened 2025-02-08
schema: support parsing and formatting of cedarschema files

masp:
*Issue #, if available:*
Partial support for https://github.com/cedar-policy/cedar-go/issues/2

*Description of changes:*
This commit adds support for parsing and formatting the human-readable schema format. It also adds basic structs for the JSON format as well, since we need to convert the human-readable schema to the JSON format.

This commit includes the following features:
- ✅ Parsing human-readable schema format (https://docs.cedarpolicy.com/schema/human-readable-schema.html)
- ✅ Marshalling and unmarshalling JSON schema format (https://docs.cedarpolicy.com/schema/json-schema.html)
- ✅ Converting human-readable to JSON
- 🆕 Pretty-printing human-readable schema format that preserves comments

This commit creates a new sub-package `github.com/cedar-policy/cedar-go/schema` and a few children packages such as `schema/ast` and `schema/token` for parsing the human-readable and JSON Cedar schema formats as well as extracting information from the AST. The AST package doesn't really need to be exposed. In theory, you could use the JSON converted format for pretty much everything, but I left it public for now.

This commit does NOT include:
- ❌ Schema validation beyond parsing syntax errors (e.g. reference checking)
- ❌ Validation of policies using the schema
- ❌ Schema annotations ([RFC48](https://github.com/cedar-policy/rfcs/blob/main/text/0048-schema-annotations.md))

After this PR, I would like to add support for a `schema/validation` package that can:
1. Validate the schema is well-formed (is logically consistent)
2. Validate a set of policies match a given schema before evaluation/authorization (`validate` on the Cedar CLI)

# Overview
The `schema` package has the following functions right now:
- `ParseFile`
- `Parse`
- `Format`
- `Convert`

These functions, when given a human-readable schema file, will return a `*ast.Schema` that is the parsed schema AST. If you then run `Format(schema, out)` it will produce a pretty-printed version of the tree preserving comments and ordering. If you use `Convert(schema)` it will return the JSON version of the schema that can be easily passed to `json.Marshal` to do the actual marshalling.

### Parsing
The parser is based on the grammar described at https://docs.cedarpolicy.com/schema/human-readable-schema-grammar.html. 

I tried my best to verify with the Rust implementation by generating a full JSON example with the Rust CLI and using it as a test case, but it's not automated so the Rust and Go implementations have deviations. As for my testing, I could find no deviations except in the Cedar playground which seems to have drifted from the Cedar CLI. AFAIK there are no test cases we can use to guarantee parity. When the validation using the schema is implemented, it can use the [test suite for the RFC](https://github.com/cedar-policy/cedar/blob/main/cedar-policy-validator/src/schema/test_579.rs).

### Formatting
The schema grammar is very simple and so pretty-printing is also very simple right now.

The rules are:
1. All declarations start on new lines
2. All lists are on a single line UNLESS it's a record type then it is forced to be over many lines.
4. Allow header, footer, and inline comments. If someone writes a comment somewhere on the tree, try to preserve where they wrote it. Since Cedar does not support /* */ the logic is simple.
5. All optional whitespace is removed (no empty lines)
6. Single element sets are reduced to just the element ([E] -> E)
7. Empty blocks ({}) are always on separate lines
8. Commas are always included after each element in a block like record types

Here's what the formatted file looks like from the tests:

```cedar
// Test case to make sure that all types of Cedar elements are parseable
// and that the formatter doesn't mess anything up.
namespace PhotoFlash {
    // Comment explaining entities User and User2
    // it's a two line comment
    entity User, User2 in UserGroup = {
        "department": String,
        "jobLevel": Long,
    } tags String;
    entity UserGroup; // inline comment entity
    entity Album in Album = { // inline comment rec type
        "account": Account,
        // record attribute comment
        "private": Bool, // record inline comment
        // footer comment rec type
        // multiple lines
    }; // footer comment entity
    entity Account = {
        "admins"?: Set<User>, // attribute comment
        "owner": User,
    };
    entity Photo in Album = {
        "account": Account,
        "private": Bool,
    };
    // type header comment
    type LongAlias = Long; // type footer comment
    action "uploadPhoto" appliesTo { // inline action comment
        // principal comment before
        principal: User, // principal comment inline
        // resource comment before
        // multi line
        resource: [Album, Photo], // resource comment inline
        // context comment before
        context: { // context comment inline
            "authenticated": Bool,
            "photo": {
                "file_size": LongAlias,
                "file_type": String,
            },
        }, // context comment after
        // remaining comments
    }; // action footer comment
    action "read"; // action footer comment 1-line
    action "all";
    action "viewPhoto", viewPhoto2 in ["read", PhotoFlash::Action::all] appliesTo {
        principal: User,
        resource: Photo,
        context: { // opener comment
            // Attribute comment (line 1)
            // Attribute comment (line 2)
            "authenticated": Bool, // attribute comment inline
        }, // context comment
    };
    action "listAlbums" in "read" appliesTo {
        principal: User,
        resource: Account,
        context: {
            "authenticated": Bool,
        },
    };
}
// Tailing comments after namespace

```

### Conversion
The human readable schema has significantly more structure than the JSON format, so `Convert` is a lossy conversion process. The JSON schema does not guarantee any ordering or comments, and everything is reduced to the minimal JSON representation. I confirmed behavior has parity in output by using the `cedar translate-schema --direction human-to-json` command. It correctly handles empty namespace elements, common types, entities, and actions. This feature should be complete.

### Testing
The schema format is fairly simple, so I tried to keep all the testing as automated as possible. I have unit tests verifying that the example human and JSON schemas are correct from the Cedar manual. I also have a test verifying that converting the human readable to JSON matches the behavior found in the Rust package. The parser also goes over some large real example schemas I found online to make sure everything works in practice.

I also added a fuzz test for the lexer + parser to verify that it won't crash easily on bad inputs. There can definitely be more rigorous testing done, but hopefully this is a good enough start.

Since this is my first PR here, let me know if I need to change or fix anything.

Thank you for your time,
Mark


patjakdev:
Thank you for your contribution, @masp. The maintainers will take a look at your PR this week. We're very excited for it!

patjakdev:
@masp, in the meantime, there are a handful of `golangci-lint` warnings and errors that should be addressed. Apologies that we don't have an easier way for you to find those during local development at this time.

masp:
Thanks Patrick, those should be fixed now.

patjakdev (changes requested):
Overall, a great PR! Thank you so much for all of this work. We are especially impressed with (and inspired by!) the level of testing of the parser. There are some requests we have before we merge this, however.

1. While we haven't been able to integrate this requirement into CI yet, we do require 100% code coverage from our tests. Since we haven't yet been able to run any of the more formal proof checking tools that the Cedar Rust library has, we figure the least we can do is say that every line of code in the module is exercised. I suspect this will take some work.

2. We would really like to avoid exposing too much of the parsing machinery publicly. Ideally, in the public package, for now we'd just have a `type Schema struct{}` with four methods: `MarshalJSON()`, `UnmarshalJSON()`, `MarshalCedar()`, and `UnmarshalCedar()`. Eventually, we'll have a `Validator` that takes a `Schema` and a `PolicySet` and validates them. We can worry about programmatic construction of `Schema`s later, but we'll probably do something like the builder pattern we used for `Policy`.

3. The pretty printer is cute, but I suspect it's overkill in terms of functionality. As far as I can tell, the Cedar Rust implementation doesn't even define a canonical pretty printed format for schemas at this time. I'd just stick with a two space indentation format and remove the flexibility around choosing the indent.

4. I suspect we're going to want a suite of higher-level types that are a bit more ergonomic to work with for validation purposes than the types you've made in the `ast` package, which are very much focused on the textual Cedar representation of the Schema. It would be great if we could re-use some of the types from the `types` package as well. However, I think we can address this concern after merging.

We'd love to collaborate in a higher-bandwidth and lower-latency manner. I've sent you an invitation to the #cedar-go Slack channel where I think we can collaborate more effectively. We're also down to chat face-to-face if you'd like; just let us know and we can follow up over email or in the aforementioned Slack channel.

patjakdev:
One other thing. In order to appease Amazon's DCO check, you're going to need to follow the instructions in the failing check to add the `Signed-Off-By:` line to each of your commits.

masp:
Thank you for the review and feedback, Patrick. I'll be happy to make those fixes.

masp:
Looks like the slack invite you sent me didn't work? Honestly not sure what went wrong. When I joined, it created a workspace and said I'd need to pay in 90 days 😅 I think I probably need to be invited to [cedar-policy.slack.com](http://cedar-policy.slack.com/)?

patjakdev:
> Looks like the slack invite you sent me didn't work? Honestly not sure what went wrong. When I joined, it created a workspace and said I'd need to pay in 90 days 😅 I think I probably need to be invited to [cedar-policy.slack.com](http://cedar-policy.slack.com/)?

@masp Well that's a bummer. The official Cedar website has this URL to join the Slack workspace: https://communityinviter.com/apps/cedar-policy/cedar-policy-language . I hope that works for you.

Once you're in, just head to the #cedar-go channel.

schema/schema.go
```diff
@@ -0,0 +1,63 @@
...
+
+// Format will pretty-print the AST node to out using the options provided.
+func Format(node ast.Node, out io.Writer, opts *FormatOptions) error {
+	return format.Node(node, out, opts)
+}
+
+// Convert will convert the AST schema to a JSON schema. The conversion process loses all comments and formatting info.
+func Convert(schema *ast.Schema) (ast.JsonSchema, error) {
+	return convert.Schema(schema)
+}
```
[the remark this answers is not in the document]
masp: I think that sounds reasonable and good first pass.

In my personal project, I want to be able to traverse the schema to generate some placeholders based off the schema fields, and so what I'm doing now is marshalling to the JSON format and walking that. Maybe some kind of walk function would be useful for the schema format as well (independent of human vs. JSON).

I'll push this to later, though.

masp:
Thanks for your patience!

I did my best to apply all the feedback you requested and I believe the result is much better. I was able to get all the code I wrote to 100% coverage, but unfortunately, I wasn't able to bring the test coverage of the generated lexer code very high (see [cedarschema.go](internal/schema/parser/cedarschema.go)). I tried for a few hours, but couldn't get all the cases since it seems to have separate branches for each character 😂 Hopefully that's mitigated enough with the fuzzer but let me know.

The new interface is now:
- MarshalCedar
- UnmarshalCedar
- MarshalJSON
- UnmarshalJSON

philhassey:
Thanks @masp!  I've put some time down for @patjakdev and myself to take a look.

schema.go
```diff
@@ -0,0 +1,68 @@
...
+// Marshalling and unmarshalling between the formats is allowed, except from a JSON schema into a human-readable one.
+type Schema struct {
+	jsonSchema  ast.JsonSchema
+	humanSchema *ast.Schema
+}
+
+// NewSchema creates an empty schema.
+func NewSchema() *Schema {
+	return &Schema{}
+}
```
patjakdev, line 25: I'd get rid of this. For the time being, it doesn't seem to provide any value over a simple `var s Schema` declaration.

I imagine that when we support programmatic construction of a schema, it will use a builder pattern similar to `Policy`.

schema.go
```diff
@@ -0,0 +1,68 @@
...
+
+// MarshalJSON serializes the schema into the JSON format.
+//
+// If the schema was loaded from UnmarshalCedar, it will convert the human-readable format into the JSON format.
+// An error is returned if the schema is invalid.
+func (s *Schema) MarshalJSON() (out []byte, err error) {
+	if s.humanSchema != nil {
+		// Error should not be possible since s.humanSchema comes from our parser.
+		// If it happens, we return empty JSON.
```
patjakdev, line 61: Seems like it would be better to return the error (or panic). Or maybe the right thing to do is to remove the possibility of an error being returned by `ast.Convert` since it's no longer callable via the public API?

schema.go
```diff
@@ -0,0 +1,68 @@
...
+	s.humanSchema, err = parser.ParseFile(filename, src, nil)
+	return
+}
+
+// MarshalCedar serializes the schema into the human readable format. This function can only be called on schemas
+// that are initialized with UnmarshalCedar, not with UnmarshalJSON.
+func (s *Schema) MarshalCedar() ([]byte, error) {
+	if s.jsonSchema != nil {
+		return nil, fmt.Errorf("cannot call MarshalCedar after UnmarshalJSON")
+	}
```
patjakdev, line 40: An unfortunate limitation, but maybe something we can live with for a bit. Keeping this code in the x/exp/schema package for a bit would give us a chance to rectify this limitation.

internal/schema/ast/convert.go
```diff
@@ -0,0 +1,137 @@
...
+
+import (
+	"fmt"
+	"strings"
+)
+
+// Convert converts an AST schema to a JSON schema. The conversion process is lossy.
+// Any information related to ordering, formatting, comments, etc... are lost completely.
+//
+// TODO: Add errors if the schema is invalid (references names that don't exist)
```
patjakdev, line 11: What's the consequence of not doing this?

internal/schema/ast/walk_test.go
```diff
@@ -0,0 +1,194 @@
...
+func TestAstScope(t *testing.T) {
+	// For each node, we verify that all of its children nodes are entirely within the Start and End of the functions.
+	// This is a quick sanity test that we didn't implement any of our start or end positions incorrectly (or not set them!)
+	tests := []string{
+		"testdata/format/test.cedarschema",
+		"testdata/format/emptynamespace.cedarschema",
+		"testdata/format/nocomments.cedarschema",
+		"testdata/walk/example.cedarschema",
+		"testdata/walk/emptyfile.cedarschema",
+	}
```
patjakdev, line 22: Please embed these with `//go:embed`

schema.go
```diff
@@ -0,0 +1,68 @@
...
+// NewSchema creates an empty schema.
+func NewSchema() *Schema {
+	return &Schema{}
+}
+
+// UnmarshalCedar parses and stores the human-readable schema from src and returns an error if the schema is invalid.
+//
+// Any errors returned will have file positions matching filename.
+func (s *Schema) UnmarshalCedar(filename string, src []byte) (err error) {
+	s.humanSchema, err = parser.ParseFile(filename, src, nil)
```
patjakdev, line 31: Make sure to clear `s.jsonSchema`

I prefer generally for functions like this that are actually constructors in disguise to do something more like this:

```
func (s *Schema) UnmarshalCedar(src []byte) error {
    hs, err := parser.ParseFile(filename, src, nil)
    if err != nil {
        return err
    }
    *s = Schema { humanSchema: hs }
}
```

That way, the "constructed" object is guaranteed to be completely overwritten by the new value.

schema.go
```diff
@@ -0,0 +1,68 @@
...
+
+// NewSchema creates an empty schema.
+func NewSchema() *Schema {
+	return &Schema{}
+}
+
+// UnmarshalCedar parses and stores the human-readable schema from src and returns an error if the schema is invalid.
+//
+// Any errors returned will have file positions matching filename.
+func (s *Schema) UnmarshalCedar(filename string, src []byte) (err error) {
```
patjakdev, line 30: Our other public `UnmarshalCedar()` functions don't take a `filename`, so this is a bit incongruous.

Maybe it could be modeled via the options pattern?

schema.go
```diff
@@ -0,0 +1,68 @@
...
+	if s.humanSchema == nil {
+		return nil, nil
+	}
+	var buf bytes.Buffer
+	err := ast.Format(s.humanSchema, &buf)
+	return buf.Bytes(), err
+}
+
+// UnmarshalJSON deserializes the JSON schema from src or returns an error if the JSON is not valid schema JSON.
+func (s *Schema) UnmarshalJSON(src []byte) error {
```
patjakdev, line 50: Same commentary about clearing `s.humanSchema`

philhassey:
Hi @masp,

Thanks for all the work on this, my main concern is that the corpus tests include a lot of schemas that are not parseable.

Add the middle bit to `corupus_test.go`

```
diff --git a/corpus_test.go b/corpus_test.go
index 613461c..d8948d2 100644
--- a/corpus_test.go
+++ b/corpus_test.go
@@ -137,6 +137,18 @@ func TestCorpus(t *testing.T) {
                                t.Fatal("error unmarshalling test", err)
                        }
 
+                       {
+                               schemaContent, err := fdm.GetFileData(tt.Schema)
+                               if err != nil {
+                                       t.Fatal("error reading schema content", err)
+                               }
+
+                               s := cedar.NewSchema()
+                               if err := s.UnmarshalCedar("test.schema", schemaContent); err != nil {
+                                       t.Fatal("error parsing schema", err, "\n===\n", string(schemaContent))
+                               }
+                       }
+
                        entitiesContent, err := fdm.GetFileData(tt.Entities)
                        if err != nil {
                                t.Fatal("error reading entities content", err)
```

Which results in a bunch of errors similar to this:

```
    --- FAIL: TestCorpus/corpus-tests/303d3f360ea13e534d8a68072ec4c0f20c1cb12b.json (0.00s)
        corpus_test.go:148: error parsing schema <input>:0:0: invalid string
            <input>:0:0: invalid string
            <input>:0:0: expected name (identifier or string)
            test.schema:1:51: unrecognized token
            test.schema:1:51: expected :, got INVALID
            test.schema:1:74: unrecognized token
            test.schema:1:75: unrecognized token
            test.schema:1:99: unrecognized token
            test.schema:1:102: unrecognized token
            test.schema:1:105: unrecognized token
            test.schema:1:106: unrecognized token
            test.schema:1:200: unrecognized token
            test.schema:1:203: unrecognized token
            test.schema:1:206: unrecognized token
            test.schema:2:1: expected , or }
            test.schema:6:2: expected , or }
            test.schema:8:1: expected }, got EOF 
            ===
             entity A in [K] = {"M"?: {""?: __cedar::String, "\0"?: __cedar::String, "\0m"?: __cedar::String, "\u{1d}\0"?: __cedar::String, "ax"?: __cedar::String, "c"?: __cedar::String, "m"?: __cedar::String, "m\u{1c}`a"?: __cedar::String, "mm"?: __cedar::String, "u"?: __cedar::String, "xc"?: __cedar::String}};
            entity K;
            action "action" appliesTo {
              principal: [A],
              resource: [A],
              context: {}
            };
```

Also, I think I'd move the Schema struct from the top package into a `x/exp/schema` package for the time being, since there isn't full coverage yet.

I hope that helps!
-Phil

patjakdev (changes requested):
Thank you again for all the hard work.

I had a few comments about the shape of the code in the public bits. The most ambiguous one is around the `filename` argument to `UnmarshalCedar()`. I'm not really sure the best way to handle that. I looked for inspiration in the Rust codebase and they just implement `FromStr` on the actual datatypes and sideload the filename into errors [at a higher level](https://github.com/cedar-policy/cedar/blob/deed11570bcbf74efc148e3d10dcc1d76257070d/cedar-policy-cli/src/lib.rs#L1459), if present.

I think my bias is just to leave filename off `UnmarshalCedar` for the time being. I'm open to being convinced the other way, though.

schema.go
masp: Ack on overwriting. With the `SetFilename`  approach, though, doesn't that make it not work? It seems like an existing bug in `policy.go` if you run `SetFilename` and then run `UnmarshalCedar` the output will not use the existing filename, right? I'll just carry over the filename if it's present.

internal/schema/ast/convert.go
masp: Only that using unresolved names (like referencing made up namespaces) will not fail.

I think it's better for this code to leave that for later passes, though, since it lets you format schemas that still have these errors, combine schemas (pseudo-imports), etc... For example, the error for schemas referencing unknown names can be returned when you pass it to the policy system to be used for validation, for example.

masp:
Thanks @philhassey! Looks like I'm not handling unicode escape characters. Great find!

Thanks again, @patjakdev, I'll have a fix shortly for those issues.

schema.go
masp: Sounds good, I don't know why I had it in my head it would be hard to implement the inverse. The limitation is removed in latest commits. It now supports going both ways.

masp:
With the latest commits, the tests in `corpus_test.go` are now being parsed successfully with the Rust supported unicode escapes. I also did some of the minor fixes mentioned, and included the conversion from JSON to human readable schema. 

A nice benefit is it formats the human readable schema much better than the Rust version:

Rust version
```cedarschema
namespace PhotoFlash {type LongAlias = __cedar::Long;
entity User in [UserGroup] = { "department": __cedar::String,
"jobLevel": __cedar::Long,
 };
entity Photo in [Album] = { "account": Account,
"private": __cedar::Bool,
 };
entity User2 in [UserGroup] = { "department": __cedar::String,
"jobLevel": __cedar::Long,
 };
entity UserGroup  = {  };
entity Account  = { "admins"?: Set < User >,
"owner": User,
 };
entity Album in [Album] = { "account": Account,
"private": __cedar::Bool,
 };
action "uploadPhoto" in [Action::"groupAction1"]appliesTo {  principal: [User], 
  resource: [Album], 
  context: { "authenticated": __cedar::Bool,
"photo": { "file_size": __cedar::Long,
"file_type": __cedar::String,
 },
 }
};
action "viewPhoto" in [random::nested::name::"actionGroup", Action::"groupAction1", Action::"groupAction2"]appliesTo {  principal: [User], 
  resource: [Photo], 
  context: { "authenticated": __cedar::Bool,
 }
};
action "groupAction1" ;
action "listAlbums" appliesTo {  principal: [User], 
  resource: [Account], 
  context: { "authenticated": __cedar::Bool,
 }
};
action "groupAction2" ;
}type Anon = __cedar::String;
action "AnonAct" ;
```

Go version:
```cedarschema
type Anon = String;
action "AnonAct";
namespace PhotoFlash {
  type LongAlias = Long;
  entity Photo in Album {
    "account": Account,
    "private": Bool,
  };
  entity User in UserGroup {
    "department": String,
    "jobLevel": Long,
  } tags String;
  entity User2 in UserGroup {
    "department": String,
    "jobLevel": Long,
  } tags String;
  entity UserGroup;
  entity Account {
    "admins"?: Set<User>,
    "owner": User,
  };
  entity Album in Album {
    "account": Account,
    "private": Bool,
  };
  action "uploadPhoto" in "groupAction1" appliesTo {
    principal: User,
    resource: Album,
    context: {
      "authenticated": Bool,
      "photo": {
        "file_size": Long,
        "file_type": String,
      },
    },
  };
  action "viewPhoto" in ["groupAction1", "groupAction2", random::nested::name::"actionGroup"] appliesTo {
    principal: User,
    resource: Photo,
    context: {
      "authenticated": Bool,
    },
  };
  action "groupAction1";
  action "groupAction2";
  action "listAlbums" appliesTo {
    principal: User,
    resource: Account,
    context: {
      "authenticated": Bool,
    },
  };
}

```

philhassey:
@masp I've put some time down this week for @patjakdev and myself to do another review.  Thanks for the changes!

schema.go
```diff
@@ -0,0 +1,79 @@
...
+// If the schema was loaded from UnmarshalCedar, it will convert the human-readable format into the JSON format.
+// An error is returned if the schema is invalid.
+func (s *Schema) MarshalJSON() (out []byte, err error) {
+	if s.humanSchema != nil {
+		// Error should not be possible since s.humanSchema comes from our parser.
+		// If it happens, we return empty JSON.
+		s.jsonSchema = ast.ConvertHuman2Json(s.humanSchema)
+	}
+	if s.jsonSchema == nil {
+		return nil, nil
```
patjakdev, line 71: `nil` is not a valid JSON encoding of an empty schema. You could write a test that marshals an empty schema then unmarshals it to prove the point.

You're probably better off just returning an error in this case.

schema.go
```diff
@@ -0,0 +1,79 @@
...
+// UnmarshalCedar parses and stores the human-readable schema from src and returns an error if the schema is invalid.
+//
+// Any errors returned will have file positions matching filename.
+func (old *Schema) UnmarshalCedar(src []byte) (err error) {
+	var s Schema
+	s.humanSchema, err = parser.ParseFile(old.filename, src)
+	if err != nil {
+		return err
+	}
+	*old = s
```
patjakdev, line 31: Seems like you need to preserve `old.filename` here...

patjakdev (approved):
Thank you again @masp for your diligence on this. Just a couple more nits, but otherwise I think this PR is ready to rock.

Please move schema.go to `x/exp/schema` to give us a bit of time to wrench on it before we commit to a finalized public shape.

masp:
Thanks a lot Patrick for all the help with the many reviews. Got those final changes in.

masp:
Hi folks, any update here?

patjakdev:
I'll work on getting a release out now.

patjakdev:
Well, nuts. Apparently something is wrong with the CI in our PR merge flow because the corpus tests are now broken.

I'll see about getting them fixed.

PR #81, opened 2025-02-21
Addresses cedar-policy/rfcs#94 concerns about time wrapping

apg:
The Go time package will silently ignore dates that don't exist on the calendar, instead carrying forward the erroneous time component. As an example, `2024-02-30` will become `2024-03-01`, since there was no February 30th, in 2024 (or ever).

This change returns an error when there's a discrepancy between the numbers we parse from the string, and the component after creating a Go time.Time value from the parsed numbers.

*Issue https://github.com/cedar-policy/rfcs/issues/94*

### DIFF
--- a/types/datetime.go
+++ b/types/datetime.go
@@ -200,6 +200,18 @@ func ParseDatetime(s string) (Datetime, error) {
 	t := time.Date(year, time.Month(month), day,
 		hour, minute, second,
 		int(time.Duration(milli)*time.Millisecond), time.UTC)
+
+	// Don't allow wrapping: https://github.com/cedar-policy/rfcs/pull/94
+	tyear, tmonth, tday := t.Date()
+	if year != tyear || time.Month(month) != tmonth || day != tday {
+		return Datetime{}, fmt.Errorf("%w: date would wrap", errDatetime)
+	}
+
+	thour, tminute, tsecond := t.Hour(), t.Minute(), t.Second()
+	if hour != thour || minute != tminute || second != tsecond {
+		return Datetime{}, fmt.Errorf("%w: time would wrap", errDatetime)
+	}
+
 	t = t.Add(offset)
 	return Datetime{value: t.UnixMilli()}, nil
 }
--- a/types/datetime.go
+++ b/types/datetime.go
@@ -77,6 +77,9 @@ func ParseDatetime(s string) (Datetime, error) {
 		return Datetime{}, fmt.Errorf("%w: invalid month", errDatetime)
 	}
 	month = 10*int(rune(s[5])-'0') + int(rune(s[6])-'0')
+	if month > 12 {
+		return Datetime{}, fmt.Errorf("%w: month is out of range", errDatetime)
+	}
 
 	if s[7] != '-' {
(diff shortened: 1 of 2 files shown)


types/datetime.go
```diff
@@ -200,6 +200,18 @@ func ParseDatetime(s string) (Datetime, error) {
...
+
+	// Don't allow wrapping: https://github.com/cedar-policy/rfcs/pull/94
+	tyear, tmonth, tday := t.Date()
+	if year != tyear || time.Month(month) != tmonth || day != tday {
+		return Datetime{}, fmt.Errorf("%w: date would wrap", errDatetime)
+	}
+
+	thour, tminute, tsecond := t.Hour(), t.Minute(), t.Second()
+	if hour != thour || minute != tminute || second != tsecond {
+		return Datetime{}, fmt.Errorf("%w: time would wrap", errDatetime)
```
adpaco-aws, line 212: The concerns on time are not about wrapping, but invalid ranges. We could just check that the values are within the associated ranges and throw a more precise error message.

adpaco-aws (commented):
The restrictions on timezone offsets are missing. I'd recommend adding a simple function to check that a value is within a given range, and use it each time after calculating the value. For example:

```go
second = 10*int(rune(s[17])-'0') + int(rune(s[18])-'0')
if !valueWithinRange(second, MAX_SECOND) {
  return Datetime{}, fmt.Errorf("%w: invalid range for seconds", errDatetime)
}
```

types/datetime.go
apg: There is a concern about time not wrapping, otherwise 2024-02-31 would pass validation under the proposal. 

I’ll change this to make more precise error messages.

philhassey:
Is this ready for review now?

apg:
@philhassey I think it can be reviewed. I should have expressed that intent by assigning reviewers. :)

philhassey (approved):
LGTM

types/datetime_test.go
```diff
@@ -100,6 +100,20 @@ func TestDatetime(t *testing.T) {
 			{"1995-01-01T00:00:00.000-0aaa", "error parsing datetime value: invalid time zone offset"},
 			{"1995-01-01T00:00:00.000-aaaa", "error parsing datetime value: invalid time zone offset"},
 			{"1995-01-01T00:00:00.000-aaaa0", "error parsing datetime value: unexpected trailer after time zone designator"},
+
+			// Prevent Wrapping invalid dates to real dates: See: cedar-policy/rfcs#94
+			{"2024-02-30T00:00:00Z", "error parsing datetime value: invalid date"},
+			{"2024-02-29T23:59:60Z", "error parsing datetime value: second is out of range"},
```
patjakdev, line 106: nit: we should actually have a positive test for the leap year case as well

patjakdev: Also, is this testing something different from the case below?

patjakdev (approved):
LGTM, just some nits about the tests.

types/datetime_test.go
apg: Actually, yes. Both are invalid because we don't support rolling over time. However, if we did roll over time:

* Line 106 would not be valid because February 30th is not a valid date.
* Line 107 would be valid because Feburary 29th is a valid date in 2024.

adpaco-aws (approved):

PR #84, opened 2025-04-08
Add support for isEmpty operator

patjakdev:
*Issue #, if available:* https://github.com/cedar-policy/cedar-go/issues/82

*Description of changes:*

`.isEmpty()` was added to the Cedar language outside of the normal RFC process. See https://github.com/cedar-policy/cedar/issues/1356 for context.

Support for `.isEmpty()` is required in order to update the corpus tests to the latest upstream version. Now that they pass, they've been updated as a part of this PR.


internal/eval/evalers.go
```diff
@@ -1035,6 +1051,7 @@ func (n *isEval) Eval(env Env) (types.Value, error) {
 	return types.Boolean(lhs.Type == n.rhs), nil
 }
 
+// isEmptyEval
```
philhassey, line 1054: This looks like it can be removed.

internal/json/json.go
```diff
@@ -97,7 +97,7 @@ type nodeJSON struct {
 	Not    *unaryJSON `json:"!,omitempty"`
 	Negate *unaryJSON `json:"neg,omitempty"`
 
-	// Binary operators: ==, !=, in, <, <=, >, >=, &&, ||, +, -, *, contains, containsAll, containsAny
+	// Binary operators: ==, !=, in, <, <=, >, >=, &&, ||, +, -, *, contains, containsAll, containsAny, isEmpty, getTag, hasTag
```
philhassey, line 100: isEmpty is a unary node, so shouldn't be in this list.  Nice that you added getTag/hasTag though!

internal/parser/cedar_unmarshal.go
```diff
@@ -682,8 +682,12 @@ func (p *parser) member() (ast.Node, error) {
 	for {
 		var ok bool
 		res, ok, err = p.access(res)
+		if err != nil {
```
philhassey, line 685: Did you add a test that proves this chance was required?  (It sure looks like a good change.)

internal/parser/node.go
```diff
@@ -140,23 +140,28 @@ func (n NodeTypeExtensionCall) precedenceLevel() nodePrecedenceLevel {
 	return accessPrecedence
 }
 
-type ContainsNode struct{}
+type SetEvaluationNode struct{}
```
philhassey, line 143: Does this need to be public?

x/exp/ast/node.go
```diff
@@ -162,6 +162,8 @@ type NodeTypeContainsAny struct {
 	BinaryNode
 }
 
+type NodeTypeIsEmpty struct{ UnaryNode }
```
philhassey, line 165: Seems a linebreak would be nice here.

internal/parser/cedar_unmarshal.go
patjakdev: Yes, one of the isEmpty tests I wrote failed to return an error when it should have, which is how I came across this bug.

internal/parser/node.go
patjakdev: It probably doesn't need to exist - we could just use `NodeTypeAccess` for all of the node types that share the `accessPrecedence` precedence level.

I also see now that `NodeTypeHasTag` has the wrong precedence...

patjakdev: Actually, no, that doesn't work since `NodeTypeAccess` wants a string argument. I suppose we could invent something like `accessPrecedenceNode` that's shared by all of the nodes with that precedence level.

PR #85, opened 2025-04-17
Add support for generic policy iteration when authorizing a request

patjakdev:
*Issue #, if available:*

*Description of changes:* Adds a new top-level `Authorize()` function which accepts a generic iterator of policies.

This functionality supersedes the existing `PolicySet.IsAuthorized()` method (which is now implemented in terms of `Authorize()`) and allows for more exotic policy injection a la the `EntityGetter` interface. This means that policy can be lazily loaded from an external source and passed to the `Authorize()` function via the iterator or policy sets can be combined for authorization without having to `Clone()` them into a new `PolicySet`.

`Authorize()` can only be called from packages built using Go 1.23 or later.

Also, several other types have gotten some iterator love:
* `types.EntityUIDSet` gets an `All()` method that returns an iterator over the entity UIDs
* `types.Set` gets an `All()` method that returns an iterator over the `Value`s in the set
* `types.Record` gets three new methods that return iterators: `All()`, `Keys()`, and `Values()` which behave similarly to `maps.All()`, `maps.Keys()`, and `maps.Values()`.
* `types.PolicySet` gets an `All()` method that returns an iterator over the `PolicyID`s and `*Policy`s in the `PolicySet`.


.github/workflows/build_and_test.yml
```diff
@@ -1,7 +1,4 @@
-# This workflow will build a golang project
-# For more information see: https://docs.github.com/en/actions/automating-builds-and-tests/building-and-testing-go
-
-name: Go
+name: Verify
```
patjakdev, line 1: I combined `build_and_test.yml` and `golangci-lint.yml` (we weren't getting much value out of the parallelism anyway) and switched to using `go-version-file` so that we only have to update the Go version in one place.

ast/ast_test.go
```diff
@@ -337,6 +337,11 @@ func TestASTByTable(t *testing.T) {
 			ast.Permit().When(ast.Long(42).ContainsAny(ast.Long(43))),
 			internalast.Permit().When(internalast.Long(42).ContainsAny(internalast.Long(43))),
 		},
+		{
+			"opContainsIsEmpty",
```
patjakdev, line 341: This is an unrelated change that I made in order to get test coverage back up to 100%

x/exp/ast/ast_test.go
```diff
@@ -451,6 +451,12 @@ func TestASTByTable(t *testing.T) {
 			ast.Policy{Effect: ast.EffectPermit, Principal: ast.ScopeTypeAll{}, Action: ast.ScopeTypeAll{}, Resource: ast.ScopeTypeAll{},
 				Conditions: []ast.ConditionType{{Condition: ast.ConditionWhen, Body: ast.NodeTypeHasTag{BinaryNode: ast.BinaryNode{Left: ast.NodeValue{Value: types.Long(42)}, Right: ast.NodeValue{Value: types.String("key")}}}}}},
 		},
+		{
+			"opIsEmpty",
```
patjakdev, line 455: More test coverage improvements

x/exp/ast/internal_test.go
```diff
@@ -9,20 +9,20 @@ import (
...
 func TestIsNode(t *testing.T) {
 	t.Parallel()
-	ScopeNode{}.isScope()
 
-	StrOpNode{}.isNode()
 	BinaryNode{}.isNode()
+	NodeTypeExtensionCall{}.isNode()
 	NodeTypeIfThenElse{}.isNode()
-	NodeTypeLike{}.isNode()
+	NodeTypeIsEmpty{}.isNode()
```
patjakdev, line 16: More test improvements.

ast/policy_test.go
```diff
@@ -12,8 +12,9 @@ import (
 func TestPolicy_MarshalJSON(t *testing.T) {
 	t.Parallel()
 
-	p := ast.Permit().PrincipalEq(types.NewEntityUID("Foo::Bar", "Baz"))
-	expected := `{
+	t.Run("roundtrip", func(t *testing.T) {
```
patjakdev, line 15: More unrelated changes to get test coverage to 100%

README.md
```diff
@@ -144,6 +144,8 @@ If you're looking to integrate Cedar into a production system, please be sure th
 
 ### New features in 1.2.0
 - Support for the .isEmpty() operator.
+- A new top-level Authorize() function, which allows authorization against a generic policy iterator instead of requiring a PolicySet. Like the EntityGetter interface does for entities, using a generic iterator enables policy to be retrieved from external sources or for policy to be selected dynamically by the iterator implementation without having to clone an entire PolicySet. Requires Go 1.23.
```
philhassey, line 147: I think not just these features but the entirety of Cedar-go requires 1.23 to compile.

authorize.go
```diff
@@ -1,28 +1,15 @@
...
-	Deny  = types.Deny
-)
-
-//revive:enable:exported
-
-// IsAuthorized uses the combination of the PolicySet and Entities to determine
+// Authorize uses the combination of the PolicySet and Entities to determine
 // if the given Request to determine Decision and Diagnostic.
-func (p PolicySet) IsAuthorized(entities types.EntityGetter, req Request) (Decision, Diagnostic) {
+func Authorize(policies iter.Seq2[PolicyID, *Policy], entities types.EntityGetter, req Request) (Decision, Diagnostic) {
```
philhassey, line 12: The other option would be to pass in an interface with a `.All()` method on it.  I wonder which is a more ergonomic or idiomatic choice.

internal/mapset/immutable.go
```diff
@@ -22,10 +23,17 @@ func (h ImmutableMapSet[T]) Intersects(o Container[T]) bool {
 
 // Iterate the items in the set, calling callback for each item. If the callback returns false, iteration is halted.
 // Iteration order is undefined.
+//
+// Deprecated: Use All() instead.
```
philhassey, line 27: I imagine these are publicly exposed via types and so that's why the Deprecated is necessary?

policy_set.go
```diff
@@ -127,3 +128,22 @@ func (p *PolicySet) UnmarshalJSON(b []byte) error {
 	}
 	return nil
 }
+
+// IsAuthorized uses the combination of the PolicySet and Entities to determine
+// if the given Request to determine Decision and Diagnostic.
+//
+// Deprecated: Use the Authorize() function instead
```
philhassey, line 135: I think even further deprecation means supplying a simple map type that has the All method on it.  Likely identical to the existing PolicySet but without the struct wrapper.  

Then the PolicySet struct wrapper can just delegate to that inside of it.

types/record_test.go
```diff
@@ -156,6 +158,139 @@ func TestRecord(t *testing.T) {
...
+			keys   mapset.ImmutableMapSet[types.String]
+			values mapset.ImmutableMapSet[types.Value]
+		}{
+			{name: "empty map", rm: types.RecordMap{}, keys: mapset.Immutable[types.String](), values: mapset.Immutable[types.Value]()},
+			{name: "one item", rm: types.RecordMap{"foo": types.Long(42)}},
+			{name: "two items", rm: types.RecordMap{"foo": types.Long(42), "bar": types.Long(1337)}},
+		}
+
+		for _, tt := range tests {
+			t.Run(tt.name+"/All", func(t *testing.T) {
```
philhassey, line 176: tt.name + "/" etc pattern is kind of weird.  Just make them sub-tests of the sub test.

types/record_test.go
```diff
@@ -156,6 +158,139 @@ func TestRecord(t *testing.T) {
 		}
 	})
 
+	t.Run("IteratorsEntire", func(t *testing.T) {
+		t.Parallel()
+
+		tests := []struct {
+			name   string
+			rm     types.RecordMap
+			keys   mapset.ImmutableMapSet[types.String]
```
philhassey, line 167: The .keys and .values are not used in the tests.  Maybe they should be or they should be removed.

types/record_test.go
```diff
@@ -156,6 +158,139 @@ func TestRecord(t *testing.T) {
...
+					i++
+					got.Add(k)
+				}
+				testutil.Equals(t, got.Len(), tt.breakOn)
+				for v := range got.All() {
+					testutil.Equals(t, slices.Contains(slices.Collect(maps.Values(rm)), v), true)
+				}
+			})
+
+			t.Run(tt.name+"/Values", func(t *testing.T) {
```
philhassey, line 282: Two tests with the same name.

types/set_test.go
```diff
@@ -136,6 +136,77 @@ func TestSet(t *testing.T) {
...
+				var got []types.Value
+				for v := range set.All() {
+					got = append(got, v)
+				}
+
+				testutil.Equals(t, len(got), len(tt.values))
+				for _, v := range got {
+					testutil.FatalIf(
+						t,
+						!slices.ContainsFunc(tt.values, func(vv types.Value) bool { return vv.Equal(v) }),
```
philhassey, line 165: I think you can use the slices.Contains generic version that doesn't require using a contains function.

policy_set.go
patjakdev: I think we could add `All()` to `PolicyMap`.

README.md
```diff
@@ -139,21 +139,31 @@ If you're looking to integrate Cedar into a production system, please be sure th
...
 - Concrete types may be replaced with compatible interfaces to expand the variety of arguments a function or method can take.
+- Backwards compatibility is maintained for all Go minor versions released within 6 months of a release of cedar-go.
 
 ## Change log
 
-### New features in 1.2.0
+### 1.2.0
+#### New Features
 - Support for the .isEmpty() operator.
+- A new top-level Authorize() function, which allows authorization against a generic policy iterator (`AuthorizationPolicySet`) instead of requiring a PolicySet. Like the EntityGetter interface does for entities, using a generic iterator enables policy to be retrieved from external sources or for policy to be selected dynamically by the iterator implementation without having to clone an entire PolicySet.
```
philhassey, line 149: I wonder if we should rename this to PolicyIterator

philhassey: Or PolicyContainer

authorize.go
```diff
@@ -1,28 +1,21 @@
...
-type DiagnosticError = types.DiagnosticError
-
-const (
-	Allow = types.Allow
-	Deny  = types.Deny
-)
-
-//revive:enable:exported
+// AuthorizationPolicySet is an interface which abstracts an iterable set of policies.
+type AuthorizationPolicySet interface {
```
philhassey, line 11: Similar to EntityGetter I'd imagine this would exist in the types package.  I kind of wish EntityGetter was called EntityContainer ...

philhassey: But I also see the signature doesn't allow for that at present.

policy_set.go
```diff
@@ -127,3 +133,22 @@ func (p *PolicySet) UnmarshalJSON(b []byte) error {
...
+func (p PolicySet) IsAuthorized(entities types.EntityGetter, req Request) (Decision, Diagnostic) {
```
philhassey, line 141: Maybe this should be a pointer receiver then?

x/exp/ast/node.go
```diff
@@ -21,19 +21,19 @@ type StrOpNode struct {
 	Value types.String
 }
 
-func (n StrOpNode) isNode() {}
+func (n StrOpNode) isNode() { _ = 0 } // No-op statement injected for code coverage instrumentation{}
```
philhassey, line 24: Oh wow, so just calling them isn't adequate?  There needs to be code in here?

policy_set.go
```diff
@@ -19,6 +20,11 @@ type PolicyID = types.PolicyID
 // PolicyMap is a map of policy IDs to policy
 type PolicyMap map[PolicyID]*Policy
 
+// All returns an iterator over the policy IDs and policies in the PolicyMap.
+func (p PolicyMap) All() iter.Seq2[PolicyID, *Policy] {
```
philhassey, line 24: I actually think that PolicyMap can be deprecated now.  It really could just be an internal implementation detail of our PolicySet object.  The 

```
func (p *PolicySet) Map() PolicyMap {
```

Similarly is no longer necessary and could be deprecated now that the PolicySet can be iterated.

policy_set.go
patjakdev: I think that changing the receiver type is considered a breaking change.

x/exp/ast/node.go
patjakdev: Yeah :\

policy_set.go
patjakdev: I think it could still be useful if you want to build a dead simple map of policies to pass to `Authorize()`

authorize.go
patjakdev: Maybe I'm missing something, but I think types should exist in the `types` package only if they need to be referenced by an `internal` package. End users should never need to refer to the `types` because the types will always be projected through the `cedar` package.

policy_set.go
patjakdev: We're going to cross our fingers that we don't break anyone. Let's also change `Get()` to take a pointer receiever.

policy_set.go
patjakdev: I'll note to use `maps.Collect(ps.All())`

PR #86, opened 2025-04-17
cedar-go: upgrade golangci-lint checks to v2 and fix all issues found

patjakdev:
*Issue #, if available:*

*Description of changes:* I ran `golangci-lint run ./... --fix` and it made all of these changes. At a cursory glance, they all seem to make sense.

I changed the config to enable the same set of linters but turned on almost all the `revive` rules. I disabled `exported` for the `internal` and `x` directories, which seems legit to me, but also had to disable it for the `types` directory for now until we can get our documentation up to snuff.

### DIFF
--- a/.github/workflows/golangci-lint.yml
+++ b/.github/workflows/golangci-lint.yml
@@ -19,34 +19,4 @@ jobs:
           go-version: "1.22"
           cache: false
       - name: golangci-lint
-        uses: golangci/golangci-lint-action@v4
-        with:
-          # Require: The version of golangci-lint to use.
-          # When `install-mode` is `binary` (default) the value can be v1.2 or v1.2.3 or `latest` to use the latest version.
-          # When `install-mode` is `goinstall` the value can be v1.2.3, `latest`, or the hash of a commit.
-          version: v1.56.2
-
-          # Optional: working directory, useful for monorepos
-          # working-directory: somedir
-
-          # Optional: golangci-lint command line arguments.
-          #
-          # Note: By default, the `.golangci.yml` file should be at the root of the repository.
-          # The location of the configuration file can be changed by using `--config=`
-          # args: --timeout=30m --config=/my/path/.golangci.yml --issues-exit-code=0
-
-          # Optional: show only new issues if it's a pull request. The default value is `false`.
-          # only-new-issues: true
-
-          # Optional: if set to true, then all caching functionality will be completely disabled,
-          #           takes precedence over all other caching options.
-          # skip-cache: true
-
-          # Optional: if set to true, then the action won't cache or restore ~/go/pkg.
-          # skip-pkg-cache: true
-
-          # Optional: if set to true, then the action won't cache or restore ~/.cache/go-build.
-          # skip-build-cache: true
-
-          # Optional: The mode to install golangci-lint. It can be 'binary' or 'goinstall'.
-          # install-mode: "goinstall"
+        uses: golangci/golangci-lint-action@v7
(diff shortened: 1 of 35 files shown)


.github/workflows/golangci-lint.yml
```diff
@@ -19,34 +19,4 @@ jobs:
...
-          # skip-cache: true
-
-          # Optional: if set to true, then the action won't cache or restore ~/go/pkg.
-          # skip-pkg-cache: true
-
-          # Optional: if set to true, then the action won't cache or restore ~/.cache/go-build.
-          # skip-build-cache: true
-
-          # Optional: The mode to install golangci-lint. It can be 'binary' or 'goinstall'.
-          # install-mode: "goinstall"
```
patjakdev, line 52: I didn't see any sense in pasting the documentation from the golangci-lint website here.

.golangci.yml
patjakdev, line 1: I decided to just use the default configuration rather than trying to upgrade our current configuration. This enabled a bunch of linters that were previously disabled, but none of them really seem objectionable to me.

.golangci.yml
```diff
@@ -1,219 +1,33 @@
...
-  - dev
-  timeout: 10m
-  modules-download-mode: vendor
-  go: 1.22
-
+version: "2"
 linters:
-  disable-all: true
   enable:
-  - bodyclose
```
patjakdev, line 11: bodyclose doesn't seem relevant for `cedar-go`

.golangci.yml
```diff
@@ -1,219 +1,33 @@
...
-  - errcheck
-  - errname
-  - gofmt
-  - goimports
-  - govet
-  - ineffassign
-  - staticcheck
-  - unused
```
patjakdev, line 19: unused doesn't appear to be a linter any more

corpus_test.go
```diff
@@ -98,7 +98,7 @@ func TestCorpus(t *testing.T) {
 	if err != nil {
 		t.Fatal("error reading corpus compressed archive header", err)
 	}
-	defer gzipReader.Close()
+	defer func() { _ = gzipReader.Close() }()
```
apg, line 101: you could also `//nolint:errcheck` this, which is more likely to survive in the future when someone sees that and is like _huh_?

internal/eval/evalers_test.go
```diff
@@ -1864,7 +1863,7 @@ func TestEntityIn(t *testing.T) {
 	}
 	t.Run("exponentialWithoutCaching", func(t *testing.T) {
 		t.Parallel(
-		// This test will run for a very long time (O(2^100)) if there isn't caching.
+			// This test will run for a very long time (O(2^100)) if there isn't caching.
```
apg, line 1866: I'm not sure why this comment is in the argument list of a nullary function, and I'm only noticing it because it was reindented. I'd advocate for putting it _below_ t.Parallel() as the t.Parallel() has nothing to do with the comment. Though, even then, putting a comment in the nullary argument list is odd.

apg (approved):
There are a few comments that I'll defer to you on, as they don't meaningfully change anything. Otherwise, seems good to me.

corpus_test.go
patjakdev: Good idea

internal/eval/evalers_test.go
patjakdev: You know, I've seen this happen in the web repo too and had to fix it there. I wonder if there's some code formatter with a bug that's somehow causing this.

PR #94, opened 2025-06-21
expose Eval function for  custom partial eval

jaredzhou:
*Issue #, if available:*
#91 
*Description of changes:*

PR #95, opened 2025-07-01
cedar-go: ignore var-naming revive lint errors in the types package

philhassey:
Addresses #94

A change in the linters is blocking a contributor's PR.

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -33,8 +33,11 @@ linters:
           - revive
         text: exported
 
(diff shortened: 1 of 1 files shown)


patjakdev (commented):
We can be more specific about disabling just the package name checks like this:

```
  settings:
    revive:
      rules:
        - name: var-naming
          arguments:
            - []
            - []
            - - skip-package-name-checks: true
 ```

patjakdev (approved):

PR #97, opened 2025-07-07
expose partial eval functions

jaredzhou:
expose Variable and Ignore in cedar-go, keep third party partial eval identical in unknown defines. and expose PartialPolicyToNode for further usage

### DIFF
--- a/internal/eval/partial.go
+++ b/internal/eval/partial.go
@@ -43,6 +43,17 @@ func IsIgnore(v types.Value) bool {
 	return false
 }
 
+// PartialPolicyToNode returns node compiled from a partially evaluated version of the policy
+// and a boolean indicating if the policy should be kept.
+// (Policies that are determined to evaluate to false are not kept.)
+func PartialPolicyToNode(env Env, p *ast.Policy) (node ast.Node, keep bool) {
+	pp, keep := PartialPolicy(env, p)
+	if !keep {
+		return ast.NewNode(nil), keep
+	}
+	return policyToNode(pp), true
+}
+
 // PartialPolicy returns a partially evaluated version of the policy and a boolean indicating if the policy should be kept.
 // (Policies that are determined to evaluate to false are not kept.)
 func PartialPolicy(env Env, p *ast.Policy) (policy *ast.Policy, keep bool) {
--- a/x/exp/eval/eval.go
+++ b/x/exp/eval/eval.go
@@ -15,3 +15,29 @@ func Eval(n ast.IsNode, env Env) (types.Value, error) {
 	evaler := eval.ToEval(n)
 	return evaler.Eval(env)
 }
+
+// PartialPolicyToNode returns node compiled from a partially evaluated version of the policy
+// and a boolean indicating if the policy should be kept.
+// (Policies that are determined to evaluate to false are not kept.)
+var PartialPolicyToNode = eval.PartialPolicyToNode
+
+// Variable is a variable in the policy.
+var Variable = eval.Variable
+
+// Ignore is a special value that is used to ignore a value.
+var Ignore = eval.Ignore
+
+// IsVariable checks if a value is a variable.
+var IsVariable = eval.IsVariable
+
+// IsIgnore checks if a value is an ignore value.
+var IsIgnore = eval.IsIgnore
+
+// ToVariable converts a value to a variable.
+var ToVariable = eval.ToVariable
+
+// TypeName returns the type name of a value.
+var TypeName = eval.TypeName
+
+// ErrType is the error type for type errors.
+var ErrType = eval.ErrType


philhassey:
Can you explain the use cases for this a bit more?  Thanks!

jaredzhou:
ok,  when i need to use cedar to list and page resource,  instead of eval all the resources, i want to generate a sql query condition. since cedar-go already has PartialPolicy function to partial eval and fold policy with already known values,  so i can `PartialPolicy` and `pocliyToNode`, and then  translate the result node to sql with function
`ToSql(node ast.IsNode, env eval.Env, mapper FieldMapper) (sql string, args []interface{}, err error)`,  now  having sql and args, i can easily combine a select sql and page resource list from db. 
below is an code example
```go
package main

import (
	"encoding/json"
	"fmt"
	"strings"

	"github.com/cedar-policy/cedar-go"
	"github.com/cedar-policy/cedar-go/x/exp/ast"
	"github.com/cedar-policy/cedar-go/x/exp/eval"
	sqlizer "github.com/jaredzhou/cedar-sqlizer"
)

func main() {
	var p = `
	permit(
		principal,
		action == Action::Files::"list",
		resource
	)
	when {
		principal.role == "member"
	}
	when {
		resource.owner == principal  || resource.is_public == true
	};
	`
	var policy cedar.Policy
	err := policy.UnmarshalCedar([]byte(p))
	if err != nil {
		panic(err)
	}
	astPolicy := (*ast.Policy)(policy.AST())

	var es = `
		[
			{
				"uid": {"type": "User", "id": "jared"},
				"parents": [],
				"attrs": {
					"role": "member"
				}
			}
		]
	`
	var em cedar.EntityMap
	err = json.Unmarshal([]byte(es), &em)
	if err != nil {
		panic(err)
	}
	env := eval.Env{
		Entities:  em,
		Principal: cedar.NewEntityUID("User", "jared"),
		Action:    cedar.NewEntityUID("Action::Files", "list"),
		Resource:  eval.Variable("resource"),
		Context:   eval.Variable("context"),
	}
	node, keep := eval.PartialPolicyToNode(env, astPolicy)
	fmt.Println(node)
	fmt.Println(keep)
	sql, args, err := sqlizer.ToSql(node.AsIsNode(), env, fileMapper{})
	fmt.Println(sql)
	fmt.Println(args)

	selectSql := fmt.Sprintf(`SELECT * from files where %s`, sql)
	fmt.Println(selectSql)
}

type fileMapper struct {
}

func (m fileMapper) Map(name string) (string, bool) {
	if strings.HasPrefix(name, "resource.") {
		return strings.Replace(name, "resource.", "file.", 1), true
	}
	return name, true
}
```
the result is 
```text
keep true
(file.owner = ? OR file.is_public = ?)
[jared true]
SELECT * from files where (file.owner = ? OR file.is_public = ?) offset 0 limit 10
```

x/exp/eval/eval.go
```diff
@@ -15,3 +15,35 @@ func Eval(n ast.IsNode, env Env) (types.Value, error) {
 	evaler := eval.ToEval(n)
 	return evaler.Eval(env)
 }
+
+// PartialPolicyToNode returns node compiled from a partially evaluated version of the policy
+// and a boolean indicating if the policy should be kept.
+// (Policies that are determined to evaluate to false are not kept.)
+var PartialPolicyToNode = eval.PartialPolicyToNode
```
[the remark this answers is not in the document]
philhassey: If you can write up (in the docs for PartialPolicy) a plausible sounding use-case for Ignore, then I think it would be fine to leave it in.

I think I would actually prefer if you just expose `PartialPolicy` as-is and then also expose `PolicyToNode` as-is.  Then in your own code you can combine them as you did (so `PartialPolicyToNode` wouldn't be part of your PR anymore.)

x/exp/eval/eval.go
```diff
@@ -15,3 +15,69 @@ func Eval(n ast.IsNode, env Env) (types.Value, error) {
...
+//
+// when the node is kept, it can be one of three kinds:
+// 1. it is a `ValueNode`, and Must be `ast.True()` (e.g. `ast.True()`)
+// 2. it is a `Node` contains `Variable` (e.g. `ast.Permit().When(ast.Context().Access("key").Equal(ast.Long(42)))`)
+// 3. it is a `Node` contains `PartialError` (e.g. `ast.ExtensionCall(partialErrorName, ast.String("type error: expected comparable value, got string"))`)
+//
+// you can use the partial evaluation result `ast.Node` to do any additional work you want
+// for example, you can convert it to an sql query.
+// in which case the variable should be a column name and binary node should be an sql expression.
+
```
philhassey, line 42: Extra linebreak here.

x/exp/eval/eval.go
```diff
@@ -15,3 +15,69 @@ func Eval(n ast.IsNode, env Env) (types.Value, error) {
...
+func IsVariable(v types.Value) bool {
+	return eval.IsVariable(v)
+}
+
+// ToVariable converts a value to a variable.
+func ToVariable(ent types.EntityUID) (types.String, bool) {
+	return eval.ToVariable(ent)
+}
+
+// TypeName returns the type name of a value.
```
philhassey, line 77: All the functions here need coverage.  Since they are well tested in another package, just a single test for each would be adequate to prove they delegate correctly.

x/exp/eval/eval.go
```diff
@@ -15,3 +15,69 @@ func Eval(n ast.IsNode, env Env) (types.Value, error) {
...
+func IsPartialError(n ast.IsNode) bool {
+	return eval.IsPartialError(n)
+}
+
+// Variable is a variable in the policy.
+func Variable(v types.String) types.Value {
+	return eval.Variable(v)
+}
+
+// IsVariable checks if a value is a variable.
```
philhassey, line 67: I think it would be really nice if these two functions were combined into a single
```
ToVariable(v types.Value) (types.String, bool)
```

Just to keep the interface smaller.  Does that seem reasonable?

x/exp/eval/eval.go
```diff
@@ -15,3 +15,69 @@ func Eval(n ast.IsNode, env Env) (types.Value, error) {
...
+	return eval.PolicyToNode(p)
+}
+
+// PartialError returns a node that represents a partial error.
+func PartialError(err error) ast.IsNode {
+	return eval.PartialError(err)
+}
+
+// IsPartialError returns true if the node is a partial error.
+func IsPartialError(n ast.IsNode) bool {
```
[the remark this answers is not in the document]
jaredzhou: yes, it's reasonable  to just unwrap the PartialError Node here

x/exp/eval/eval.go
jaredzhou: good idea, i was just exposing exist function,  yes marshal and unmarsh looks more elegant

philhassey (changes requested):
This looks good now, there is still a tiny bit of coverage missing then I think it'll be ready.  I'll have @patjakdev take a look as well.  Thanks!

image

jaredzhou:
yes, i forgot to `git add eval_test.go` , now it's ready.  thanks for reviewing.

philhassey (approved):
LGTM

patjakdev (dismissed):
Just one thing I'm concerned about.

internal/eval/partial.go
```diff
@@ -508,6 +508,32 @@ func extError(err error) ast.NodeTypeExtensionCall {
...
+		return nil, false
+	}
+	if ec.Name != partialErrorName {
+		return nil, false
+	}
+	if len(ec.Args) != 1 {
+		return nil, false
+	}
+	evaler := ToEval(ec.Args[0])
+	v, _ := evaler.Eval(Env{})
```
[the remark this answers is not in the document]
jaredzhou: it could fail with some handed constructed Node , so i will just handle the error

PR #101, opened 2025-09-19
Use standard io Writer and Reader where possible

sagikazarmark:
*Description of changes:*

This PR contains a few refactors and features:

Refactor: use standard `io.Writer` and `io.Reader` where possible instead of `bytes.Buffer` and `[]byte`. This is necessary for the feature in the PR, but it's also for good hygiene (relying on standard interfaces).

Feature: add a Decoder to the parser package (and x/exp) that allows reading multiple policies from any `io.Reader` (practically: files). It is modeled after the decoder in the stdlib JSON package.

The primary use case is not having to read all contents of a file first, but let the component that actually needs the content do the reading (which is good practice).

Naturally, an encoder would similarly make sense, but I wanted to open a PR first to get feedback.

### DIFF
--- a/internal/parser/cedar_unmarshal.go
+++ b/internal/parser/cedar_unmarshal.go
@@ -36,39 +36,35 @@ func (p *PolicySlice) UnmarshalCedar(b []byte) error {
 }
 
 type Decoder struct {
-	r io.Reader
-
-	once sync.Once
-
-	parser *parser
+	createParser func() (*parser, error)
 }
 
 func NewDecoder(r io.Reader) *Decoder {
-	return &Decoder{r: r}
+	return &Decoder{
+		createParser: sync.OnceValues(func() (*parser, error) {
+			tokens, err := TokenizeReader(r)
+			if err != nil {
+				return nil, err
+			}
+
+			p := newParser(tokens)
+
+			return &p, nil
+		}),
+	}
 }
 
 func (d *Decoder) Decode(p *Policy) error {
-	var err error
-
-	d.once.Do(func() {
-		tokens, e := TokenizeReader(d.r)
-		if e != nil {
-			err = e
-			return
-		}
-
-		parser := newParser(tokens)
-		d.parser = &parser
-	})
+	parser, err := d.createParser()
 	if err != nil {
 		return err
 	}
 
-	if d.parser.peek().isEOF() {
+	if parser.peek().isEOF() {
 		return io.EOF
 	}
 
-	return p.fromCedar(d.parser)
+	return p.fromCedar(parser)
 }
 
 func (p *Policy) UnmarshalCedar(b []byte) error {
--- a/internal/parser/cedar_unmarshal_test.go
+++ b/internal/parser/cedar_unmarshal_test.go
@@ -579,6 +579,7 @@ func TestDecoderErrors(t *testing.T) {
 			decoder := parser.NewDecoder(strings.NewReader(tt))
 
 			testutil.Error(t, decoder.Decode(nil))
+			testutil.Error(t, decoder.Decode(nil))
 		})
 	}
 }


internal/parser/cedar_marshal.go
```diff
@@ -1,29 +1,29 @@
...
 	"github.com/cedar-policy/cedar-go/internal/extensions"
 	"github.com/cedar-policy/cedar-go/x/exp/ast"
 )
 
-func (p *Policy) MarshalCedar(buf *bytes.Buffer) {
+func (p *Policy) MarshalCedar(w io.Writer) {
 	for _, a := range p.Annotations {
-		marshalAnnotation(a, buf)
-		buf.WriteRune('\n')
+		marshalAnnotation(a, w)
```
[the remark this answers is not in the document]
sagikazarmark: This is the marshaling part which would be used in an encoder which I haven't added (yet).

I think relying on the standard interfaces is a better solution: less room for error.

As for using bytes.Buffer on the decoder/encoder side: I want to use files directly :)

sagikazarmark:
Coverage should be a 100% now.

internal/parser/cedar_marshal.go
```diff
@@ -24,6 +27,22 @@ func (p *Policy) MarshalCedar(buf *bytes.Buffer) {
 	}
 
 	buf.WriteRune(';')
+
+	_, _ = w.Write(buf.Bytes())
+}
+
+type Encoder struct {
```
philhassey, line 34: I think you should move this Encoder into the `ext/parser` package and in there you could call the `func (p *Policy) MarshalCedar(buf *bytes.Buffer)` method in this package.  Then you could write the buffer to the writer in `Encode` and return the error there, again avoiding changing anything in `internal/parser`.  I think that'll get you want you need, and avoid unnecessary changes to the internal/parser package.

internal/parser/cedar_unmarshal.go
```diff
@@ -33,6 +35,42 @@ func (p *PolicySlice) UnmarshalCedar(b []byte) error {
 	return nil
 }
 
+type Decoder struct {
```
philhassey, line 38: Can we move this whole decoder struct into the `exp/parser` package?

sagikazarmark: No, we can't. It uses unexported features of the parser package.

sagikazarmark:
@philhassey I've just realized I've been working with the wrong `Policy` type the whole time. 🤦 

Obviously, encoding/decoding `parser.Policy` is necessary, but far from enough since that's not the exported `Policy`. All these policy structs confused me. 😄 

Answering your latest round of review:

- No, the decoder can't be moved out of parser, because it relies on exported features
- As a result, I think the encoder should remain in the parser package too
- The exp/parser package is not even necessary, because it's not what as I user I want to interact with
- So I kept the decoder and encoder in the parser package as they are
- Reverted changes to existing marshaling/unmarshaling code as you requested (with the exception that the Tokenizer accepts a reader as well)
- I did **NOT** add doc comments to the decoder and encoder in the parser package, because none of the other functions have them

What changed with the latest version of the PR:

I added an encoder and a decoder type to the root package.

Since they follow standard Go practices, I think their API is stable enough to skip the exp package (though I'm happy to move them if you give me guidance where exactly they should be placed)

They call into the parser encoder/decoder (which is still necessary due to how parsing works internally)

The public encoder/decoder have doc comments as requested.

Sorry for the confusion. But I think this PR is now solid.

philhassey (dismissed):
Okay, I think the shape looks good to me now.  I'll get one more maintainer to give it a look just to be sure.  Thanks a ton for the work on this!  There's a little bit of coverage still missing:

image

I do wonder if in that function it'd be better if it used OnceValue to return the error so that the error was returned repeatedly in all future calls?  Otherwise I think it might panic in line 67.  

So maybe a test to call it a second time to prove it correctly emits the same error again in that case?

sagikazarmark:
Good catch @philhassey !

Fixed coverage and added another call Decode.

I believe I replied to both of your comments. If I left anything unanswered, let me know.

internal/parser/cedar_unmarshal.go
```diff
@@ -33,6 +35,38 @@ func (p *PolicySlice) UnmarshalCedar(b []byte) error {
...
+	createParser func() (*parser, error)
+}
+
+func NewDecoder(r io.Reader) *Decoder {
```
philhassey, line 42: @patjakdev @sagikazarmark How's this for a variation, I think it handles the errors how I want, does what @sagikazarmark and avoids using sync like @patjakdev wants:

```
type Decoder struct {
	r         io.Reader
	parser    *parser
	parserErr error
}

func (d *Decoder) getParser() (*parser, error) {
	if d.parser != nil || d.parserErr != nil {
		return d.parser, d.parserErr
	}
	tokens, err := TokenizeReader(d.r)
	if err != nil {
		d.parserErr = err
		return nil, err
	}
	p := newParser(tokens)
	d.parser = &p
	return d.parser, nil
}

func NewDecoder(r io.Reader) *Decoder {
	return &Decoder{
		r: r,
	}
}

func (d *Decoder) Decode(p *Policy) error {
	parser, err := d.getParser()
	if err != nil {
		return err
	}

	if parser.peek().isEOF() {
		return io.EOF
	}

	return p.fromCedar(parser)
}
```

sagikazarmark: If the tokenizer gets rewritten anyway, I don't think it makes much of a difference (except maybe that the current solution is more idiomatic IMO).

I can make the change if you want me to, but my vote goes to the current version.

sagikazarmark:
I started working on making the tokenizer branching off from this one.

If you are good with this PR, could you merge it? I can provide a PR for the tokenizer shortly.

PR #116, opened 2025-10-28
Add DOT graph export for EntityMap

phsym:
Add function `dot.Write(w io.Writer, entities iter.Seq[types.Entity]) error` to export a **DOT** graph of the entities.

This function is inspired from [Entities::to_dot_str](https://docs.rs/cedar-policy-core/4.7.0/cedar_policy_core/entities/struct.Entities.html#method.to_dot_str) method from the Rust implementation

### DIFF
--- a/x/exp/dot/dot.go
+++ b/x/exp/dot/dot.go
@@ -1,0 +1,76 @@
+package dot
+
+import (
+	"fmt"
+	"io"
+	"iter"
+	"strconv"
+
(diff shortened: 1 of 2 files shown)


philhassey (changes requested):
A suggestion to move this into x/exp and make it more generic.  Thanks!

phsym:
awesome, thank you

Issue #103, opened 2025-10-01
Support RFC #71 (trailing commas)


patjakdev:
https://github.com/cedar-policy/rfcs/blob/main/text/0071-trailing-commas.md

PR #123, opened 2025-11-23
add feather: trailing commas

jaredzhou:
*Issue #, if available:*
#103 
*Description of changes:*
add trailing commas for 
1. entity list
2. expr list, including set and function call arg list
3. record


jaredzhou:
ok, i will add the function args test

philhassey (approved):
Thanks!

PR #124, opened 2025-11-25
Add sumtype:decl to Declaration type

gregory-nisbet:
We also need to fix up one location `convertNamespace` which handles nested namespaces. Nested namespaces are not valid (and don't even parse), but here I'm handling them with an error rather than a panic.

This leads to some changes in functions that transitively call convertNamespace.

*Issue #, if available:*

sumtype:decl extracted from PR below (but the other code changes are not)
https://github.com/cedar-policy/cedar-go/pull/111

### DIFF
--- a/internal/schema/ast/ast.go
+++ b/internal/schema/ast/ast.go
@@ -97,6 +97,7 @@ func (s *Schema) End() token.Position {
 	return token.Position{}
 }
 
+//sumtype:decl
 type Declaration interface {
 	Node
 	isDecl()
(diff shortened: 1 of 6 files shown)


patjakdev (changes requested):
I think what the linter is trying to tell us here is that `Namespace`  is not a `Declaration`. Indeed, the grammar says as much:

```
Schema    := {Namespace}
Namespace := (Annotations 'namespace' Path '{' {Decl} '}') | Decl
Decl      := Entity | Action | TypeDecl
```

It seems like the thing to do is to narrow `Declaration` to `Entity`, `Action`, and `CommonTypeDecl` and then introduce a new sum type for `(Namespace | Decl)` that's used in `Schema`. I'm not sure exactly what to do with `CommentBlock`, but I don't think it matters too much.

patjakdev:
Closing PR due to inactivity

PR #127, opened 2026-01-29
types: add support RFC 110 expanded year format

patjakdev:
*Issue #, if available:*
https://github.com/cedar-policy/cedar-go/issues/128

*Description of changes:*

This change adds support for the nine-digit expanded year datetime format specified in [RFC 110](https://github.com/cedar-policy/rfcs/pull/110), allowing exotic datetimes far in the past or future to be represented in Cedar policies and entities.

Incidentally, it also fixes a bug in the `Datetime` type where timezone offset was being interpreted incorrectly (see https://github.com/cedar-policy/cedar-go/issues/128).

### DIFF
--- a/types/datetime.go
+++ b/types/datetime.go
@@ -12,13 +12,23 @@ import (
 
 var errDatetime = internal.ErrDatetime
 
+// maxDatetime is the highest possible timestamp that will fit in 64 bits of millisecond-precision space.
+var maxDatetime = time.Date(292278994, 8, 17, 7, 12, 55, 807*1e6, time.UTC)
+
+// minDatetime is the lowest possible timestamp that will fit in 64 bits of millisecond-precision space.
+var minDatetime = time.Date(-292275055, 5, 17, 16, 47, 04, 192*1e6, time.UTC)
+
 // Datetime represents a Cedar datetime value
 type Datetime struct {
 	// value is a timestamp in milliseconds
 	value int64
 }
 
-// NewDatetime returns a Cedar Datetime from a Go time.Time value
+// NewDatetime returns a Cedar Datetime from a Go time.Time value.
+//
+// The provided time.Time is truncated to millisecond precision. The result is
+// undefined if the Unix time in milliseconds cannot be represented by an int64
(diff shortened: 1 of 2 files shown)


types/datetime_test.go
```diff
@@ -58,66 +71,70 @@ func TestDatetime(t *testing.T) {
...
+			{"1970-01-01T00:00:00+2360", "error parsing datetime value: offset minutes is greater than 59"},
+			{"1970-01-01T00:00:00-2360", "error parsing datetime value: offset minutes is greater than 59"},
+
+			{"+", "error parsing datetime value: unexpected EOF"},
+			{"-", "error parsing datetime value: unexpected EOF"},
+			{"+12345678", "error parsing datetime value: unexpected EOF"},
+			{"+1234-01-01", "error parsing datetime value: invalid year"},
+			{"+00000000a-01-01", "error parsing datetime value: invalid year"},
+			{"-abcdefghi-01-01", "error parsing datetime value: invalid year"},
+			{"+12345678A-01-01", "error parsing datetime value: invalid year"},
```
philhassey: I think there should be an error case that is exactly 1ms too high or 1ms too low.

types/datetime.go
```diff
@@ -39,185 +77,146 @@ func NewDatetimeFromMillis(ms int64) Datetime {
...
-	if time.Month(month) != tmonth || day != tday {
-		return Datetime{}, fmt.Errorf("%w: invalid date", errDatetime)
+	if len(s) > 0 {
+		return Datetime{}, fmt.Errorf("%w: unexpected additional characters", errDatetime)
 	}
 
+	t := time.Date(year, time.Month(month), int(day),
+		int(hour), int(minute), int(second),
+		int(time.Duration(milli)*time.Millisecond), time.UTC)
+
```
philhassey: Though maybe with the addition of the offset, it's impossible to pre-validate the h:m:s

philhassey: Or maybe the validate just has to happen after the offset is added?  I'm not sure what is right here.

PR #130, opened 2026-02-10
x/exp/schema: fix reserved keyword handling and add validation

patjakdev:
*Issue #, if available:*

*Description of changes:*
Add a tokenReservedKeyword token type to the schema parser's lexer, matching the approach used in the main Cedar parser. Previously, reserved keywords like "true", "false", "in", "if", etc. were lexed as plain identifiers, which meant the parser silently accepted them in positions where they should be rejected (e.g. `entity true;`, `type if = String;`).

Bugs fixed:
- Reserved Cedar keywords were accepted as entity, type, and action names, namespace path components, and attribute names without quoting. The parser now rejects these with a clear error message.
- `__cedar` as a definition name (entity, type, enum) was silently accepted. These are now rejected while still allowing `__cedar` as an action name, attribute name, and type reference prefix, which matches the Cedar Rust behavior.
- Duplicate annotations (e.g. `@doc("a") @doc("b")`) were silently accepted with last-wins semantics. The parser now rejects duplicates.
- Duplicate `principal`, `resource`, or `context` declarations within `appliesTo` were silently accepted. The parser now rejects duplicates.
- Empty `principal` or `resource` type lists in `appliesTo` (e.g.  `principal: []`) were silently accepted, producing a meaningless empty list. The parser now rejects these.
- `appliesTo` blocks missing a `principal` or `resource` declaration were accepted. The parser now requires both.
- `MarshalSchema()` emitted reserved keywords as bare identifiers in attribute and action names (e.g. `true: String`), producing output that could not be re-parsed. `isValidIdent()` now checks for reserved keywords and the marshaler quotes them.

### DIFF
--- a/internal/parser/cedar_tokenize.go
+++ b/internal/parser/cedar_tokenize.go
@@ -36,9 +36,12 @@ type Token struct {
 	Text string
 }
 
-// N.B. "is" is included here for compatibility with the Rust implementation. The Cedar specification does not list
-// "is" as a reserved keyword
-var reservedKeywords = []string{"true", "false", "if", "then", "else", "in", "like", "has", "is"}
+var reservedKeywords = []string{"true", "false", "if", "then", "else", "in", "like", "has", "is", "__cedar"}
+
+// IsReservedKeyword reports whether s is a reserved Cedar keyword.
+func IsReservedKeyword(s string) bool {
+	return slices.Contains(reservedKeywords, s)
+}
 
 func (t Token) isEOF() bool {
 	return t.Type == TokenEOF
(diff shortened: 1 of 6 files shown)


internal/parser/cedar_tokenize.go
```diff
@@ -36,9 +36,12 @@ type Token struct {
 	Text string
 }
 
-// N.B. "is" is included here for compatibility with the Rust implementation. The Cedar specification does not list
```
patjakdev: `is` is now in the official documentation

philhassey (approved):
LGTM

PR #133, opened 2026-03-04
Fix duration parsing to support i64 range and add Makefile targets

philhassey:
- Fix ParseDuration overflow check from MaxInt32 to proper int64 overflow detection at digit accumulation, unit multiplication, and total accumulation steps.
- Add overflow checking to datetime offset and durationSince eval operations to prevent silent int64 wraparound.
- Add Makefile targets: test, linters, corpus-download, corpus-convert, corpus-update.
- Update DEVELOPMENT.md with make targets, DCO requirements.
- Update corpus GitHub action issue body to reference make targets.
- Suppress false-positive staticcheck warning on json:"-,omitempty" tag.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -9,7 +9,7 @@ linters:
 	golangci-lint run
 	go run github.com/alecthomas/go-check-sumtype/cmd/go-check-sumtype@latest -default-signifies-exhaustive=false ./...
 	go test -coverprofile=coverage.out ./...
-	sed -i'' -e '/^github.com\/cedar-policy\/cedar-go\/internal\/schema\/parser\/cedarschema.go/d' coverage.out
+	sed -i '' '/^github.com\/cedar-policy\/cedar-go\/internal\/schema\/parser\/cedarschema.go/d' coverage.out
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -1,3 +1,22 @@
...
+	go tool cover -func=coverage.out | sed 's/%$$//' | awk '{ if ($$3 < 100.0) { printf "Insufficient code coverage for %s\n", $$0; failed=1 } } END { exit failed }'
+
+# Download the latest corpus tests tarball
+corpus-tests.tar.gz:
+	curl -L -o corpus-tests.tar.gz https://raw.githubusercontent.com/cedar-policy/cedar-integration-tests/main/corpus-tests.tar.gz
+
+# Convert Cedar schemas to JSON schemas
 corpus-tests-json-schemas.tar.gz: corpus-tests.tar.gz
 	@echo "Generating JSON schemas from Cedar schemas..."
 	@rm -rf /tmp/corpus-tests /tmp/corpus-tests-json-schemas
```
patjakdev: You could use `mktemp -d` to make a unique temporary directory each time so you don't have to worry about what might have existed previously.

That will also remove the dependency on `/tmp`, which might differ from the user's `$TMPDIR` of choice.

philhassey: Good idea, done!

PR #154, opened 2026-05-23
parser: accept annotations without a value

iainmcgin:
## Summary

Makes the value of a policy annotation optional in the textual parser, so a bare `@name` is accepted as shorthand for `@name("")`.

This matches the Cedar reference grammar, which made the annotation value optional in [cedar-policy/cedar 4.2.0](https://github.com/cedar-policy/cedar/blob/main/cedar-policy/CHANGELOG.md) ("Annotations without explicit values", resolving cedar-policy/cedar#1031). Before this change, cedar-go rejected the bare form with a parse error (`exact got permit want (`).

## Details

- `internal/parser/cedar_unmarshal.go`: the `("...")` group is now optional. An absent value yields `""`, identical to `@name("")` — consistent with how Cedar collapses an absent annotation value to the empty string in its AST.
- The duplicate-annotation check moved ahead of the optional value parse, so duplicates are detected uniformly across bare and valued forms. One observable side effect: an input like `@key @key(...)` now reports `duplicate annotation: @key` rather than a syntax error on the value — a clearer message, and not a behavior change (both forms were already rejected).
- The marshaler is intentionally unchanged: an empty value still serializes as `@name("")`. That form parses on every Cedar version, so this PR is a strict superset of accepted inputs with no change to existing output.

## Testing

- New positive cases: bare annotation, bare reserved-keyword key, two distinct bare annotations, and bare-preceding-valued.
- New error cases: duplicate bare annotations, and duplicate across bare + valued forms.
- A test asserting `@foo` and `@foo("")` parse to identical ASTs, and that an empty value marshals back to `@foo("")`.
- `go test ./...`, `go vet ./...`, and `gofmt -l` are clean.

### DIFF
--- a/internal/parser/cedar_parse_test.go
+++ b/internal/parser/cedar_parse_test.go
@@ -82,6 +82,19 @@ func TestParse(t *testing.T) {
 				action,
 				resource
 			  );`, false},
+		{"bareAnnotation", `@bare
+			  permit(
+				principal,
+				action,
+				resource
+			  );`, false},
+		{"bareAndValuedAnnotations", `@bare
+			  @valued("v")
+			  permit(
+				principal,
+				action,
+				resource
+			  );`, false},
 
 		// Additional success cases
 		{"primaryInt", `permit(principal, action, resource) when { 1234 };`, false},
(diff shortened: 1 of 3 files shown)


patjakdev (approved):
Thank you for your contribution!

Issue #153, opened 2026-05-22
ParseIPAddr accepts IPv6 zone identifiers


patjakdev:
## Summary

`ip("fe80::1%eth0")` errors in cedar-rust but is accepted by cedar-go. Cedar policy semantics do not include host-scope/zone, so the zone is semantically meaningless and creates another cross-implementation divergence.

## Root cause

`ParseIPAddr()` delegates to `netip.ParseAddr` / `netip.ParsePrefix`, which accept zone suffixes per RFC 4007.

## Recommended fix

Reject zoned addresses explicitly. After `netip.ParsePrefix` / `netip.ParseAddr` succeeds, check `addr.Zone() != ""` and return `errIP`:

```go
} else if net, err := netip.ParsePrefix(s); err == nil {
    if net.Addr().Zone() != "" {
        return IPAddr{}, fmt.Errorf("%w: IPv6 zone identifiers are not supported", errIP)
    }
    return IPAddr(net), nil
} else if addr, err := netip.ParseAddr(s); err == nil {
    if addr.Zone() != "" {
        return IPAddr{}, fmt.Errorf("%w: IPv6 zone identifiers are not supported", errIP)
    }
    return IPAddr(netip.PrefixFrom(addr, addr.BitLen())), nil
}
```

Add tests for `fe80::1%eth0` and `fe80::1%eth0/64`.

## Priority

Low–Medium. Easy fix; aligns Go behavior with Rust.

TheFellow:
Resolved in #156.

PR #156, opened 2026-05-29
Reject IPv6 zone identifiers in Cedar IP parsing.

TheFellow:
This change now rejects
- named link-local zone
- numeric interface zone
- CIDR-looking suffix swallowed as a zone by ParseAddr
- non-link local zoned IPv6

*Issue #, if available:*
#153 

*Description of changes:*

ParseIPAddr now returns an IP parse error for zoned IPv6 inputs like fe80::1%eth0, fe80::1%1, and fe80::1%eth0/64, matching Cedar Rust semantics. Added parser tests covering named zones, numeric zones, prefix-like zoned input, and non-link-local zoned IPv6.


types/ipaddr.go
```diff
@@ -24,6 +24,9 @@ func ParseIPAddr(s string) (IPAddr, error) {
 	} else if net, err := netip.ParsePrefix(s); err == nil {
```
TheFellow, line 24: `netip.ParsePrefix` rejects zones before returning success

patjakdev (approved):

